PR #75, opened 2022-06-28
feat: update agents for parameter tests

TomMcL:
### Description

Updating Agents to move them into a separate folder and more permanent location. Additionally adding the concept of `scenarios` which should make it easier to wrap up agents and an environment into a distinct event/trading structure.  

In future these can then hopefully be reused both in instances such as testing governance parameter changes and as tests in their own right.

Additionally reworking the waiting structure for two purposes. Firstly to abstract out whether the code is forwarding the blockchain (in the case of null chain) or sleeping for an approximate time (if it were to connect to a pre-running node). Secondly, to allow time steps specified in blocks when forwarding time. This is necessary because often what we really want is to forward a certain number of blocks, however the null chain only forwards in time increments.


witgaw (approved):
LGTM

PR #76, opened 2022-06-29
feat: Adding order stream api and tests

TomMcL:
### Description

Adding the functionality to allow the service to listen to Order events and then maintain an up-to-date view of all orders on the market.

Requires some handling around threading as we're reading/writing to the same object, however the locking should be largely covering this.

Tests added to ensure working functionality


witgaw (approved):
LGTM (Jenkins got a bit of a meltdown it seems though)

TomMcL:
@witgaw Thanks 👍 . Yeah something seems funky with Jenkins...

PR #78, opened 2022-07-05
feat: Creating a new version of the 'Ideal Market Maker' simulation designed to split out the various actors

TomMcL:
### Description

Creating a new version of Ideal Market Maker in which each agent performs separately without knowledge of the others. This differs from v1 in that v1 constructs a 'perfect' market at each step to behave according to the Market Maker's expectations. In v2, we instead have actors act purely statistically. Ideally this should result in comparable behaviour over longer numbers of steps, though is likely to show some differences on shorter timescales.


vega_sim/scenario/ideal_market_maker_v2/utils/price_process.py
```diff
@@ -0,0 +1,62 @@
...
+    # for i in range(len(time_step)-1):
+    #     S[i+1] = S[i] + sigma*dW[i]
+
+    for i in range(len(time_step) - 1):
+        S[i + 1] = S[i] * np.exp(-0.5 * sigma * sigma * dt + sigma * dW[i])
+
+    # market decimal place
+    S = np.round(S, mdp)
+
+    # make sure the simulated external price is larger than 0
```
Galen-Cao: Could we delete the code before 'while True' here?

Galen-Cao (dismissed):
LGTM, and I like the idea to simulate the LOB shape, will try and play after merged.

vega_sim/scenario/ideal_market_maker_v2/utils/price_process.py
TomMcL: Ah hadn't actually read this, hmm.

I'm more inclined to either trim the price off at zero or throw an error I think. With the current setup it's using a new random state if it hits a zero so you might end up with repeats which don't come out at the same result even though you thought you were using the same random seed. What do you reckon?

Galen-Cao: Yeah, I think it'll be better to use the same random state and throw an error if the price <= 0 as the situation of price <= 0 is rare.

TomMcL: Updated, now if a random state is specified it will error if that state doesn't generate another process. If no state is specified it will perform as before and keep trying new things until one works :+1:

Galen-Cao (approved):
LGTM

PR #88, opened 2022-07-14
Feat/parameter tests

Galen-Cao:
Change some parameters in agents:
- initial topup
- create parameter `market_name` and `asset_name` in OptimalMarketMaker class to better set the market and asset name
- fix the liquidity provision in OptimalMarketMaker

### DIFF
--- a/vega_sim/api/trading.py
+++ b/vega_sim/api/trading.py
@@ -292,7 +292,7 @@ def submit_simple_liquidity(
         delta_sell:
             int, the offset from reference point for the sell side of LP
     """
-    print(f"Submitting with buy {delta_buy} and sell {delta_sell}")
+    # print(f"Submitting liquidity with buy {delta_buy} and sell {delta_sell}")
     return submit_liquidity(
         wallet_name=wallet_name,
         wallet=wallet,
--- a/vega_sim/api/trading.py
+++ b/vega_sim/api/trading.py
@@ -292,7 +292,6 @@ def submit_simple_liquidity(
         delta_sell:
             int, the offset from reference point for the sell side of LP
     """
-    # print(f"Submitting liquidity with buy {delta_buy} and sell {delta_sell}")
     return submit_liquidity(
         wallet_name=wallet_name,
         wallet=wallet,
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -219,12 +219,12 @@ class BackgroundMarket(StateAgentWithWallet):
         )
         new_price = self.price_process[self.current_step]
         new_buy_shape = self._calculate_price_volume_levels(
-            new_price - self.spread,
+            new_price - self.spread / 2,
             new_price,
             side=vega_protos.SIDE_BUY,
         )
         new_sell_shape = self._calculate_price_volume_levels(
-            new_price + self.spread,
+            new_price + self.spread / 2,
             new_price,
             side=vega_protos.SIDE_SELL,
         )
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -30,12 +30,14 @@ class MarketOrderTrader(StateAgentWithWallet):
         wallet_pass: str,
         market_name: str,
         asset_name: str,
+        initial_asset_mint: float = 1000000,
         buy_intensity: float = 1,
         sell_intensity: float = 1,
         tag: str = "",
         random_state: Optional[np.random.RandomState] = None,
     ):
         super().__init__(wallet_name + str(tag), wallet_pass)
+        self.initial_asset_mint = initial_asset_mint
         self.buy_intensity = buy_intensity
         self.sell_intensity = sell_intensity
         self.tag = tag
@@ -61,7 +63,7 @@ class MarketOrderTrader(StateAgentWithWallet):
         self.vega.mint(
             self.wallet_name,
             asset=asset_id,
-            amount=1e15,
+            amount=self.initial_asset_mint,
         )
         self.vega.wait_fn(2)
 
@@ -116,6 +118,7 @@ class BackgroundMarket(StateAgentWithWallet):
         market_name: str,
         asset_name: str,
         price_process: List[float],
+        initial_asset_mint: float = 1000000,
         spread: float = 0.02,
         tick_spacing: float = 0.01,
         num_levels_per_side: int = 20,
@@ -125,6 +128,7 @@ class BackgroundMarket(StateAgentWithWallet):
     ):
         super().__init__(wallet_name + tag, wallet_pass)
         self.price_process = price_process
+        self.initial_asset_mint = initial_asset_mint
         self.spread = spread
         self.current_step = 0
         self.tag = tag
@@ -152,7 +156,7 @@ class BackgroundMarket(StateAgentWithWallet):
         self.vega.mint(
             self.wallet_name,
             asset=asset_id,
-            amount=1e15,
+            amount=self.initial_asset_mint,
         )
         self.vega.wait_fn(2)
 
@@ -271,11 +275,13 @@ class OpenAuctionPass(StateAgentWithWallet):
         side: str,
         market_name: str,
         asset_name: str,
+        initial_asset_mint: float = 1000000,
         initial_price: float = 0.3,
         tag: str = "",
     ):
         super().__init__(wallet_name + tag, wallet_pass)
         self.side = side
+        self.initial_asset_mint = initial_asset_mint
         self.initial_price = initial_price
         self.tag = tag
         self.market_name = market_name
@@ -297,7 +303,7 @@ class OpenAuctionPass(StateAgentWithWallet):
         self.vega.mint(
             self.wallet_name,
             asset=tDAI_id,
-            amount=1e15,
+            amount=self.initial_asset_mint,
         )
         self.vega.wait_fn(2)
 
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -56,13 +56,12 @@ class MarketOrderTrader(StateAgentWithWallet):
         ][0]
 
         # Get asset id
-
         asset_id = self.vega.find_asset_id(symbol=self.asset_name)
         # Top up asset
         self.vega.mint(
             self.wallet_name,
             asset=asset_id,
-            amount=100000,
+            amount=1e15,
         )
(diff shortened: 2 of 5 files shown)


vega_sim/api/trading.py
```diff
@@ -292,7 +292,7 @@ def submit_simple_liquidity(
         delta_sell:
             int, the offset from reference point for the sell side of LP
     """
-    print(f"Submitting with buy {delta_buy} and sell {delta_sell}")
+    # print(f"Submitting liquidity with buy {delta_buy} and sell {delta_sell}")
```
davidsiska-vega: Let's try to avoid "commented out code" in the main branch.

davidsiska-vega (changes requested):
The other stuff I leave for @TomMcL .

vega_sim/api/trading.py
TomMcL: I dislike that I even added a print at all. Kill it please @Galen-Cao :)

Galen-Cao: No problem.

vega_sim/scenario/common/agents.py
```diff
@@ -56,13 +56,12 @@ def initialise(self, vega: VegaServiceNull):
...
 
         # Get asset id
-
         asset_id = self.vega.find_asset_id(symbol=self.asset_name)
         # Top up asset
         self.vega.mint(
             self.wallet_name,
             asset=asset_id,
-            amount=100000,
+            amount=1e15,
```
TomMcL: Why does this need to be so vast? 

This should also probably be a parameter when configuring the agent (called `initial_asset_mint` or something more exciting) rather than a hardcoded value, so if you could pull out that would be great (given I assume the issue here is that you need it 1e15 for the sim)

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -53,7 +53,9 @@ def __init__(
         asset_decimal: int = 5,
         market_decimal: int = 5,
         market_position_decimal: int = 2,
-        commitment_amount: float = 6000,
+        market_name: str = None,
+        asset_name: str = None,
+        commitment_amount: float = 20000000000,
```
TomMcL: I think commitment amount 6000 is probably a more standard 'baseline' than the agent committing $20bn to the market (also slightly concerning if that ends up necessary for a 'realistic' simulation?)

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -122,21 +128,20 @@ def initialise(self, vega: VegaServiceNull):
...
+            max_faucet_amount=1e20,
         )
         self.vega.wait_fn(5)
         self.vega.wait_for_datanode_sync()
         # Get asset id
-        self.tdai_id = self.vega.find_asset_id(symbol=ccy_name)
+        self.tdai_id = self.vega.find_asset_id(symbol=self.asset_name)
         # Top up asset
-        self.initial = 100000
+        self.initial = 1e15
```
TomMcL: Similarly for this can we promote it to being passed in to the agent as a parameter?

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -300,7 +339,7 @@ def _place_or_amend_order(
                 side=side,
                 volume=volume,
                 order_type=vega_protos.Order.Type.TYPE_LIMIT,
-                wait=False,
+                wait=True,
```
TomMcL: Why are we now waiting here? Does this slow things down?

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -235,35 +264,40 @@ def step(self, vega_state: VegaState):
...
+            pass
+
+        self.vega.submit_simple_liquidity(
+            wallet_name=self.wallet_name,
+            market_id=self.market_id,
+            commitment_amount=self.commitment_amount,
+            fee=0.002,
+            reference_buy="PEGGED_REFERENCE_BEST_BID",
+            reference_sell="PEGGED_REFERENCE_BEST_ASK",
+            delta_buy=self.bid_depth - self.spread,
```
TomMcL: This should be `- spread/2` I think?

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -235,35 +264,40 @@ def step(self, vega_state: VegaState):
...
+                .values()
+            ):
+                if order.side == vega_protos.SIDE_BUY:
+                    buy_order = order
+                else:
+                    sell_order = order
+            
+        except Exception as error:
+            print(error)
+            pass
```
TomMcL: I don't think this should be throwing any errors? What's getting thrown in here (And why do we want to ignore whatever error *is* thrown?)

As a side note: If we were just logging errors we'd want to use `logger.exception` rather than print, but I suspect we don't want the try/except here at all.

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -311,4 +350,4 @@ def _place_or_amend_order(
                 pegged_reference=vega_protos.PEGGED_REFERENCE_MID,
                 pegged_offset=offset,
                 volume_delta=volume - order.size,
-            )
+            )
```
TomMcL: newline at end of file

vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -198,4 +215,8 @@ def run_iteration(self, vega: VegaServiceNull, pause_at_completion: bool = False
         run_with_console=True,
         seconds_per_block=40,  # Heuristic
     ) as vega:
-        scenario.run_iteration(vega=vega, pause_at_completion=True)
+        scenario.run_iteration(
+            vega=vega, 
+            pause_at_completion=True,
+            run_with_console=False,
```
TomMcL: run_with_console here I think shouldn't do anything because it's the outer VegaServiceNull which is controlling that.

vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -32,13 +32,15 @@ def __init__(
...
         asset_decimal: int = 5,
+        market_position_decimal: int = 2,
         initial_price: float = 100,
         sigma: float = 1,
         kappa: float = 1,
         q_upper: int = 20,
         q_lower: int = -20,
         alpha: float = 10**-4,
         phi: float = 5 * 10**-6,
+        lp_commitamount: float = 200000,
```
TomMcL, line 46: Ok 200k sounds more reasonable 👍

TomMcL (commented):
Few thoughts, and let's kill anything commented out

vega_sim/scenario/ideal_market_maker_v2/agents.py
davidsiska-vega: Another parameter?

Galen-Cao: Yeah, here is just a default value, will change it to a normal level.

vega_sim/scenario/ideal_market_maker_v2/agents.py
Galen-Cao: It got a 'Key error' when searching the orders, but I think the issue would be fixed after the latest change, will delete this.

vega_sim/scenario/ideal_market_maker_v2/agents.py
Galen-Cao: The distance of best bid/ask from mid should be spread, right (because we have the shape starts at price - spread)?       
 `new_buy_shape = self._calculate_price_volume_levels(
            new_price - self.spread,
            new_price,
            side=vega_protos.SIDE_BUY,)
`
So, the pegged order would be +- spread?

vega_sim/scenario/ideal_market_maker_v2/agents.py
Galen-Cao: Just set 'True' to make sure the order submitted successfully, most of the time will use instead amend_order function, so no big impact on the running time.

vega_sim/scenario/ideal_market_maker_v2/agents.py
TomMcL: Spread would (should) be difference between best bid and best ask, not mid-price to best X.

Galen-Cao: If so, should we also change buy/sell shape in BackgroundMarket, as the new_buy/sell_shape starts at new_price - self.spread, new_price + self.spread, where the spread is actually 2*self.spread (I assume to use the same self.spread in both agents)?

Galen-Cao: Another idea: how about not to set the self.spread for market_maker, instead to let market_maker call the information of current LOB to know the current best bid/ask and spread of the market to adjust the offset?

TomMcL: I'm not sure the market maker actually even uses the spread it is given anymore looking at it

TomMcL: But yeah I think the spread shape in BackgroundMarket should be fixed too 👍

vega_sim/scenario/ideal_market_maker_v2/agents.py
```diff
@@ -245,25 +253,29 @@ def step(self, vega_state: VegaState):
...
         )
-        self.current_step += 1
+
+        # self._place_orders(
+        #     buy_offset=self.bid_depth,
+        #     sell_offset=self.ask_depth,
+        #     volume=1,
+        #     buy_order=buy_order,
+        #     sell_order=sell_order,
+        # )
```
TomMcL: Why are we no longer placing orders here?

Galen-Cao: The market maker will place pegged order from liquidity provision, so I think we do not need to place limit order by hand here?

TomMcL: Liquidity provision is only deployed when needed, whereas placing limit orders makes it more explicit and ensures the orders are actually on the book at all times (Having a deeper LOB means the orders may not always be deployed otherwise).

davidsiska-vega (approved):

PR #90, opened 2022-07-14
feat: move parameter tests into market sim from research internal

TomMcL:
### Description

Dump of Parameter Test code from research-internal into public repo. No major changes other than renaming some folders.


Galen-Cao (commented):
I'm just wondering why not to move the latest parameter-test in research-internal, but it's totally okay to use the origin one if better.

TomMcL:
> I'm just wondering why not to move the latest parameter-test in research-internal, but it's totally okay to use the origin one if better.

This was meant to be a straight dump of the latest code (barring some moving of files about) from research-internal to here. Do you mean I used an old version/missed changes?

Galen-Cao (dismissed):
LGTE

Galen-Cao (approved):
LGTM

PR #93, opened 2022-07-18
feat: market crash

TomMcL:


witgaw (approved):
LGTM (to the extent where I can actually understand what's going on)

PR #98, opened 2022-07-20
Parameter test

Galen-Cao:
- Add logging function of LOB
- Plot LOB in ipynb file
- An update in configs


TomMcL:
None of these commits are 'conventional commits' ;). We can fix up in the squash + merge but best to start off with them that way

vega_sim/parameter_test/parameter/loggers.py
```diff
@@ -174,3 +173,43 @@ def tau_scaling_additional_data(
...
+    vega: VegaServiceNull,
+    agents: List[Agent],
+) -> Dict[str, Any]:
+    mm_agent = [
+        agent
+        for agent in agents
+        if isinstance(agent, (OptimalMarketMakerV2, OptimalMarketMaker))
+    ][0]
+
+    order_book = vega.open_orders_by_market(market_id=mm_agent.market_id)
```
TomMcL: if you use `vega.order_status_from_feed(live_only=True)` here instead you should see it run a lot faster

TomMcL: Although that said, I think `vega.market_depth` should give you what you want here without having to do the processing at all?

vega_sim/parameter_test/parameter/experiment.py
```diff
@@ -106,14 +107,35 @@ def output_logs(
...
+                csv_writer = csv.writer(f, delimiter=",")
+                # Keys should not change throughout, so just take the first
+                headers = (
+                    ['Time Step', 'Order Book Bid Side', 'Order Book Ask Side']
+                )
+
+                csv_writer.writerow(["Iteration"] + headers)
+                for i, res in enumerate(result_list):
+                    for row in res:
+                        csv_writer.writerow([i] + [row[c] for c in headers])
```
TomMcL: I feel like there's probably a cleaner way to do this than checking for the string in the header, having a think about what that might be and will get back to you!

TomMcL (commented):
Mostly looking good, a couple of queries/suggestions. Additionally:
- TauScaling looks like it's outputs aren't cleared, can you clear those before we do merge please?
- Haven't looked through notebook plotting logic/tested yet, but looks fine at a first glance
- That said, would you be able to install Poetry (python package manager) and add plotly/whatever other packages are needed for the plotting for the notebooks to the `jupyter` 'extras' section please? Give me a shout if anything with poetry is non-obvious 👍

vega_sim/parameter_test/parameter/loggers.py
Galen-Cao: Will try both methods to choose which better to use.

vega_sim/parameter_test/parameter/experiment.py
TomMcL: So, thinking the best thing here is if you:
- Pull out `file_pattern` to an optional argument which defaults to `FILE_PATTERN`
- Add a flag which defaults to `True` for whether to save the `run_config.json` file
- Add an additional field to `SingleParameterExperiment` which is a list of (named?) tuples containing `(file_pattern, column_list)`
   -  Have the default be `None` and if it's set to that stick to standard behaviour
   - If column_list is `None`, include everything

How does that sound?

Galen-Cao: Looks great to me about the first two points, will add them. Could I ask about the third point -  if we'd like to output LOB in a new file (could be done if we pass a list of (file_pattern, column_list) in `SingleParameterExperiment` ), and we will use the column_list in `output_logs`, then to check for the data in column_list, right?

TomMcL: Yeah that's what I'm thinking

TomMcL:
Also, make sure to merge with the latest changes in `develop` which I think is why the tests are failing

TomMcL (approved):
Lgtm

PR #104, opened 2022-07-28
feat: Move the sides on background market based on background price

### LINKED ISSUE #103: Update Background Market agent to move sides in correct order
Background market agent should move it's orders first to widen the spread then narrow on each step to avoid crossing (e.g. if the price is increasing move asks first, if decreasing move bids)

TomMcL:
### Description

Move background market agent's orders first to widen the spread then narrow on each step to avoid crossing (e.g. if the price is increasing move asks first, if decreasing move bids)

Closes #103


Galen-Cao (approved):
LGTM

PR #105, opened 2022-08-01
feat: merge protos into Vega and remove datanode

TomMcL:
Now that datanode and protos are both integrated into the main Vega repository we can drop the additional dependencies and run solely from that.

Also fixing a few bugs with proposal IDs and some extra logging


witgaw (approved):
LGTM 💥

TomMcL:
Merging as failing tests due to known issue which appears to be upstream in Vega core repo

PR #107, opened 2022-08-02
feat: Adding random state fixing to parameter experiments to ensure comparability

TomMcL:
### Description

Generate a set of random states (one per 'run') for scenario testing to ensure that for each parameter setting a broadly comparable chain of events is observed.


vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -145,6 +151,7 @@ def set_up_background_market(
             tag=str(tag),
             buy_intensity=self.buy_intensity,
             sell_intensity=self.sell_intensity,
+            random_state=random_state,
```
Galen-Cao, line 154: Should we also add 'random state' in MarketOrderTrader agent when posting MOs to fix randomness?

Galen-Cao (commented):
Leave a comment, please ignore if doesn't make sense

vega_sim/scenario/ideal_market_maker_v2/scenario.py
TomMcL: I think that's what this is doing here no? Or am I missing using it somewhere?

Galen-Cao: Ah, I confused it with another agent, my bad. Looks fine to me now.

PR #109, opened 2022-08-10
feat: Adding random state fixing to parameter experiments to ensure comparability

TomMcL:
### Description

Generate a set of random states (one per 'run') for scenario testing to ensure that for each parameter setting a broadly comparable chain of events is observed.


Galen-Cao (approved):
LGTM

PR #110, opened 2022-08-10
feat: remove validation time from non asset proposals

TomMcL:
### Description

Validation timestamp is only required for asset proposals. Recently a change was made in core to make it an error to pass it for others, so remove it here


Galen-Cao (approved):
LGTM

PR #113, opened 2022-08-16
feat: Updating protobuf generation to new version

TomMcL:
### Description

Updating the protobuf generation framework to use `buf` instead of the current python module version as new changes to the Vega core protobuf layout make this a more usable approach (due to some plugins). This also simplifies the call slightly.


Jiajia-Cui (approved):
LGTM

PR #117, opened 2022-08-17
feat: Pin repo to vega version and provide script to update when tests pass

### LINKED ISSUE #116: Pin market-sim to specific versions of Vega core
Currently we always run from the `develop` version of Vega core which can result in frequent breakages when breaking changes are made. It would be nice to control this more by moving to pinned versions which can be iteratively updated by a process when all tests pass. 

We can also combine this with tagging Vega-market-sim itself to create known working releases.

TomMcL:
### Description

Adding functionality to update Vega version based on commit tag, with a script to run tests and automatically update the current version when they are passing

Closes #116


cdummett (approved):
LGTM

PR #121, opened 2022-08-18
fix: Clone random states to properly replicate in parameter tests

TomMcL:
### Description

A minor fix to properly clone random states which we use to ensure the same price series/action series process is used by agents across runs where possible.

The issue was that we were reusing the *exact* random state object, which obviously just continued producing numbers from where the previous one left off. Instead we use a cloned version so that each will start fresh.


Galen-Cao (approved):
LGTM

PR #122, opened 2022-08-18
Feat/limit order trader

### LINKED ISSUE #112: Create an agent to place limit orders (semi-)randomly
In aid of #111 we need an agent who can post limit orders somewhat randomly. 

The agent should:

- Decide whether it is going to do anything at all with some probability (probably independently select whether to buy/sell)
- If it is, select a distance from the mid to place said order (this should possibly also permit negative values sometimes for completeness?)
- For each order select a 'type'. Ideally the distribution should be customisable, but we probably want the majority to be either GTC (Good Til' Cancelled) or maybe GTT (Good Til' Time) better to avoid build up
- Select a size
- Place it's order
- Keep track of it's orders and maybe randomly cancel them sometimes too

cdummett:
**Description:**
SemiRandomLimitOrderTrader agent added to common agents and implemented in ideal_market_maker_v2 scenario.

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -901,10 +901,7 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
             volume = self.sell_volume * self.random_state.poisson(self.sell_intensity)
             price = best_offer_price - (random_offset - ln_mean)
 
-        blockchain_time = self.vega.trading_data_client.GetVegaTime(
-            data_node_protos.trading_data.GetVegaTimeRequest()
-        ).timestamp
-        expires_at = int(blockchain_time + self.duration * 1e9)
+        expires_at = (self.vega.get_blockchain_time() + self.duration) * 1e9
 
         self.vega.submit_order(
             trading_wallet=self.wallet_name,
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 import numpy as np
+
+from math import exp
+
 from collections import namedtuple
 from typing import List, Optional, Union
 from numpy.typing import ArrayLike
@@ -718,14 +721,12 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
     the asset for that market, and mints itself the specified quantity of the
     required asset.
 
-    At any given step; the agent randomly choses whether to submit an order
-    using a variable order volume and a randomly sampled price following a
-    normal distribution. For the normal distribution; mu is selected as either
-    the best bid or best offer price for sell and buy orders respectively and
-    sigma is a variable parameter.
+    At any given step; the agent has an adjustable probability of submitting
+    an order and an adjustable probability of cancelling a randomly
+    selected order.
 
-    At any given step; The agent also choses whether to cancel a randomly
-    selected open order sampled from a uniform distribution.
+    When submitting an order; the agent choses a price following a lognormal
+    distribution where the underlying normal distribution can be adjusted.
 
     """
 
@@ -753,7 +754,8 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
             "TIME_IN_FORCE_IOC": 0.2,
             "TIME_IN_FORCE_FOK": 0.1,
         },
-        sigma: Optional[float] = 5.0,
+        mean: Optional[float] = 2.0,
+        sigma: Optional[float] = 1.0,
     ):
         """Init the object and class attributes.
 
@@ -786,8 +788,10 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
                 Dictionary of side options and probabilities.
             time_in_force_opts (dict, optional):
                 Dictionary of time in force options and probabilities.
-            float (float, optional):
-                Standard deviation of the normal distribution
+            mean (float, optional):
+                Mean of the log-normal distribution.
+            sigma (float, optional):
+                Standard deviation of the log-normal distribution.
         """
 
         super().__init__(wallet_name + str(tag), wallet_pass)
@@ -806,6 +810,7 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
         self.cancel_bias = cancel_bias
         self.side_opts = side_opts
         self.time_in_force_opts = time_in_force_opts
+        self.mean = mean
         self.sigma = sigma
 
     def initialise(self, vega: VegaServiceNull):
@@ -867,17 +872,21 @@ class SemiRandomLimitOrderTrader(StateAgentWithWallet):
             p=list(self.time_in_force_opts.values()),
         )
 
+        random_offset = self.random_state.lognormal(
+            mean=self.mean,
+            sigma=self.sigma,
+            )
+        ln_mean = exp(self.mean + self.sigma ** 2 / 2)
+        
         if side == "SIDE_BUY":
 
             volume = self.buy_volume
-            mu = best_bid_price
+            price = best_bid_price + (random_offset - ln_mean)
 
         elif side == "SIDE_SELL":
 
             volume = self.sell_volume
-            mu = best_offer_price
-
-        price = self.random_state.normal(loc=mu, scale=self.sigma)
+            price = best_offer_price - (random_offset - ln_mean)
 
         self.vega.submit_order(
             trading_wallet=self.wallet_name,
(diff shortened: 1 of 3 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+
+        if agent_order_side == "buy":
+
+            volume = self.random_state.poisson(self.buy_intensity)
+            side = vega_protos.SIDE_BUY
+            mu = best_bid_price
+
+        elif agent_order_side == "sell":
+
+            volume = sell_vol = self.random_state.poisson(self.sell_intensity)
```
Galen-Cao: Is there a typo of 'sell_vol' here?

cdummett: Thanks Galen, yep that is a typo

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+
+
+class SemiRandomLimitOrderTrader(StateAgentWithWallet):
+    """Agent which randomly submits and cancels limit orders at prices normally distributed about the best bid price or best offer price.
+
+    At initalisation; the agent creates a wallet, identifies the market id and the asset for that market, and mints itself the specified quantity of the required asset.
+
+    At any given step; the agent randomly choses whether to submit an order using a randomly sampled order volume following a poission distribution and a randomly sampled price following a normal distribution. For the normal distribution; mu is selected as either the best bid or best offer price for sell and buy orders respecitvely and sigma is selected as a factor of the spread.
+
+    At any given step; The agent also choses whether to cancel a randomly selected open order sampled from a uniform distribution.
```
TomMcL: Can we break these lines up?

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
             self.vega.settle_market(
                 self.terminate_wallet_name, self.settlement_price, self.market_id
             )
+
+
+class SemiRandomLimitOrderTrader(StateAgentWithWallet):
+    """Agent which randomly submits and cancels limit orders at prices normally distributed about the best bid price or best offer price.
+
+    At initalisation; the agent creates a wallet, identifies the market id and the asset for that market, and mints itself the specified quantity of the required asset.
```
TomMcL: initialisation

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+
+            agent_submit_order = self.random_state.choice([0, 1, 1])
+            if agent_submit_order:
+                self._sumbit_order()
+
+            agent_cancel_order = self.random_state.choice([0, 0, 1])
+            if agent_cancel_order:
+                self._cancel_order(vega_state=vega_state)
+
+    def _sumbit_order(self):
```
TomMcL: submit

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+            self.market_id
+        ].state == markets_protos.Market.State.STATE_ACTIVE:
+
+            agent_submit_order = self.random_state.choice([0, 1, 1])
+            if agent_submit_order:
+                self._sumbit_order()
+
+            agent_cancel_order = self.random_state.choice([0, 0, 1])
+            if agent_cancel_order:
+                self._cancel_order(vega_state=vega_state)
```
TomMcL, line 871: Could we make these customisable parameters for probability? (Could generate a random number and use `<=` / `>=`)

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+
+        best_bid_price, best_offer_price = self.vega.best_prices(
+            market_id=self.market_id
+        )
+
+        agent_order_side = self.random_state.choice(["buy", "sell"])
+
+        if agent_order_side == "buy":
+
+            volume = self.random_state.poisson(self.buy_intensity)
```
TomMcL: Can we update this to multiply by a passed in "trade unit size"? I recently realised (but haven't yet updated prod to fix) that just directly selecting from poisson means we can only do unit order sizes.

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +710,151 @@ def finalise(self):
...
+            mu = best_bid_price
+
+        elif agent_order_side == "sell":
+
+            volume = self.random_state.poisson(self.sell_intensity)
+            side = vega_protos.SIDE_SELL
+            mu = best_offer_price
+
+        sigma = self.spread * 2
+        price = self.random_state.normal(loc=mu, scale=sigma)
```
TomMcL: I feel like maybe lognormal rather than normal might be fun here, but open to persuasion. Sigma maybe should be another parameter rather than always twice the spread.

vega_sim/scenario/common/agents.py
```diff
@@ -6,6 +6,7 @@
 from typing import List, Optional, Union
 from numpy.typing import ArrayLike
 from vega_sim.api.data import Order
+from vega_sim.api.helpers import num_to_padded_int
```
TomMcL: Is this actually used anywhere? (hopefully isn't needed :) )

TomMcL (changes requested):
Looking good bar some small tweaks. Could we also have it choose the time_in_force randomly too? Not sure how we would make that customisable so can maybe stick to having it semi-hardcoded

vega_sim/scenario/common/agents.py
cdummett: Yep will remove, was used at one point but don't need anymore

vega_sim/scenario/common/agents.py
cdummett: Sure, have replaced buy and sell intensity arguments with buy and sell volume arguments

vega_sim/scenario/common/agents.py
cdummett: Sounds good, bias arguments have been added to the class for both submit and cancel order methods

TomMcL: 👍

vega_sim/scenario/common/agents.py
TomMcL: Ah, may have been unclear, keep the poisson randomness but multiply by the passed buy/sell volume (so we want both parameters :) )

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +712,206 @@ def finalise(self):
...
+        buy_volume: float = 2.5,
+        sell_volume: float = 2.5,
+        tag: str = "",
+        random_state: Optional[np.random.RandomState] = None,
+        submit_bias: float = 0.5,
+        cancel_bias: float = 0.5,
+        side_opts: Optional[dict] = {
+            "SIDE_BUY": 0.5,
+            "SIDE_SELL": 0.5,
+        },
```
TomMcL: This should default to `None` then set `foo if side_opts is None else side_opts` in the body of the `__init__`. Will explain why in person.

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +712,206 @@ def finalise(self):
...
+        )
+
+        side = self.random_state.choice(
+            a=list(self.side_opts.keys()),
+            p=list(self.side_opts.values()),
+        )
+        time_in_force = self.random_state.choice(
+            a=list(self.time_in_force_opts.keys()),
+            p=list(self.time_in_force_opts.values()),
+        )
```
TomMcL, line 886: I think in the case of GTT we'll need to actually pass a time it is good until.. somewhere. I haven't used GTT yet but assume it'll be obvious in the docs 👍

vega_sim/scenario/common/agents.py
```diff
@@ -709,3 +714,223 @@ def finalise(self):
...
+            price = best_bid_price + (random_offset - ln_mean)
+
+        elif side == "SIDE_SELL":
+
+            volume = self.sell_volume * self.random_state.poisson(self.sell_intensity)
+            price = best_offer_price - (random_offset - ln_mean)
+
+        blockchain_time = self.vega.trading_data_client.GetVegaTime(
+            data_node_protos.trading_data.GetVegaTimeRequest()
+        ).timestamp
```
TomMcL: `self.vega.get_blockchain_time()`

TomMcL:
One last small tweak and I think we're good.

Can we also remove it from the ideal_market_maker_v2 scenario for now? I'm not sure how the agent will change the parameter tests so let's play it safe and keep it out 😄

cdummett:
Sure thing thats a fair point. Have removed it 😄

vega_sim/scenario/common/agents.py
cdummett: Changed 👍

TomMcL (approved):
LGTM

PR #124, opened 2022-08-19
feat: Move to Console v2

TomMcL:
### Description

Move to using console v2 rather than current v1 (as v1 is deprecated and largely no longer working).

Console v2 currently does not fully work due to enum issues, but shouldn't be a blocker on merging this.

Users should ensure to fully delete `extern/console` and `vega_sim/bin/console` folders before running their next `make ui` to avoid version issues.


Galen-Cao (approved):
LGTM

PR #125, opened 2022-08-19
feat: Historic price loader

TomMcL:
### Description

Functionality for loading historic prices from Coinbase, alongside various quality of life tweaks, such as the ability to automatically launch a web browser pointing to the GraphQL node.


vega_sim/scenario/common/agents.py
```diff
@@ -79,14 +82,25 @@ def initialise(self, vega: VegaServiceNull):
...
+        self.vega.submit_simple_liquidity(
+            self.wallet_name,
+            self.market_id,
+            1000,
+            "0.0015",
+            vega_protos.PEGGED_REFERENCE_BEST_BID,
+            vega_protos.PEGGED_REFERENCE_BEST_ASK,
+            1,
+            1,
+        )
```
Galen-Cao: I know there is no impact on trades, but could I ask why we need the market order trader post liquidity here?

TomMcL: Ah, we definitely don't. Was using this branch for some custom simulation so missed cutting some bits out, thanks for catching :+1:

Galen-Cao (approved):
LGTM

PR #131, opened 2022-08-25
Fix/parameter test

Galen-Cao:
### Description

Update and simplify simulation of ideal_market_maker_v1 (controlled version): fix the issue of no trading with price changes in the sim; move main features from ideal_market_maker_v2; add new agent types in the sim; fix issues in notebook plot. 

### Testing

Ran current parameter test simulation

### Breaking Changes

None

### Closes

None


vega_sim/scenario/ideal_market_maker/agents.py
```diff
@@ -267,8 +291,8 @@ def step(self, vega_state: VegaState):
         self.vega.submit_simple_liquidity(
             wallet_name=self.wallet_name,
             market_id=self.market_id,
-            commitment_amount=100,
-            fee=0.002,
+            commitment_amount=self.commitamoumt,
```
TomMcL: typo

vega_sim/scenario/ideal_market_maker/agents.py
```diff
@@ -572,3 +589,277 @@ def initialise(self, vega: VegaServiceNull):
...
+            self.wallet_name,
+            asset=tDAI_id,
+            amount=self.initial_asset_mint,
+        )
+        self.vega.wait_for_total_catchup()
+
+    def step(self, vega_state: VegaState):
+        position = self.vega.positions_by_market(
+            wallet_name=self.wallet_name, market_id=self.market_id
+        )
```
TomMcL, line 814: It may not be in there currently, and not for this PR, but we should add all traders' positions to the vega_state so it doesn't have to be called individually

vega_sim/scenario/ideal_market_maker/agents.py
```diff
@@ -572,3 +589,277 @@ def initialise(self, vega: VegaServiceNull):
...
+            self.vega.submit_market_order(
+                trading_wallet=self.wallet_name,
+                market_id=self.market_id,
+                side=trade_side,
+                volume=current_position,
+                wait=True,
+                fill_or_kill=False,
+            )
+
+        Order_book = []
```
TomMcL: `order_book`

vega_sim/scenario/ideal_market_maker/agents.py
```diff
@@ -572,3 +589,277 @@ def initialise(self, vega: VegaServiceNull):
...
+        for _, orders in (
+            self.vega.order_status_from_feed(live_only=True)
+            .get(self.market_id, {})
+            .items()
+        ):
+            Order_book += list(orders.values())
```
TomMcL: I wonder whether just querying with `Vega.market_depth` would be clearer here. It's a query to datanode though so might end up slower than just calculating ourselves. 

Possibly the `VegaService` should start listening to order book updates too which would solve this.

vega_sim/scenario/ideal_market_maker/agents.py
```diff
@@ -572,3 +589,277 @@ def initialise(self, vega: VegaServiceNull):
...
+
+        if volume:
+            self.vega.submit_market_order(
+                trading_wallet=self.wallet_name,
+                market_id=self.market_id,
+                side=trade_side,
+                volume=volume,
+                wait=True,
+                fill_or_kill=False,
+            )
```
TomMcL, line 859: Could we move this to the `common` agents repository? Seems a useful one in general. 

Would also be cool to add a `proportion_taken` (or better name) config parameter to it which could multiply with the `volume` here so that it doesn't have to take *all* of the 'informed' price advantage.

vega_sim/scenario/ideal_market_maker/scenario.py
```diff
@@ -133,16 +165,37 @@ def set_up_background_market(
...
+        )
+
         env = MarketEnvironment(
             base_agents=[
                 market_maker,
                 tradingbot,
                 randomtrader,
                 auctionpass1,
                 auctionpass2,
+                # liquidityprovider,
```
TomMcL, line 198: Deliberate?

TomMcL (changes requested):
Couple of tweaks but mostly looks good to me.

Haven't been through the notebooks

TomMcL:
You may also need a merge with latest `develop`

vega_sim/scenario/ideal_market_maker/agents.py
Galen-Cao: Nice catch, thank you.

vega_sim/scenario/ideal_market_maker/agents.py
Galen-Cao: The idea of `proportion_take` is cool, will add it.

vega_sim/scenario/ideal_market_maker/scenario.py
Galen-Cao: Yep, one can easily add a liquidity provider in the sim just cancel the comment of this line. Don't you like it? If not, I can remove.

cdummett:
Not part of this PR but I am having issues running the notebooks as the 'plotly' package is missing 

(its also not in the poetry.lock or pyproject.toml files).

Adding plotly with poetry resolves the issue. Should this be included in this PR?

Galen-Cao:
> Not part of this PR but I am having issues running the notebooks as the 'plotly' package is missing

I remembered we add plotly as dependency but I cannot find that one in poetry.lock file either, I can add it again.

vega_sim/scenario/ideal_market_maker/scenario.py
TomMcL: No it's fine just checking you didn't mean to have it in 😄

TomMcL:
Intriguing, yeah it should be an extra in `pyproject.toml` but not sure where it's gone! Galen if you could add it back as an extra in the `Jupyter` section that would be great 👍

vega_sim/scenario/ideal_market_maker/agents.py
Galen-Cao: Yeah, it would be much more clearer to use `Vega.market_depth`, so change to it.

cdummett:
Notebooks are running fine for me now in a fresh poetry environment 👍

Galen-Cao:
> Notebooks are running fine for me now in a fresh poetry environment 👍

Cool, thank you for the testing. I am asking for the help from Tom to pass the Jenkins check, do you also have any advice?

TomMcL (approved):
LGTM

PR #133, opened 2022-08-25
Feat: comprehensive_market scenario added

### LINKED ISSUE #111: Run a comprehensive Scenario covering a 'long' period of time and containing as many possible actions as feasible
It would be useful to have a scenario test which runs a market and has traders/LPs exercising as many different functionalities as possible in an environment with a reasonably large number of agents and over a reasonably large number of timesteps. This should include:

- 1000+ timesteps
- Market order agents 5+ placing FoK and IoC orders
- Limit order agents 5+ selecting prices at different depths placing orders across the range of types (GTC/GTT/GFA etc)
- LPs utilising shaped orders and limit orders
- All limit agents also using pegged orders

cdummett:
### Description
Scenario comprehensive_market added to available vega_sim scenarios. Scenario simulates market order traders, limit order traders, and liquidity providers trading over a long time period. 

### Testing
None yet.

### Breaking Changes
None yet.

### Closes
#111


vega_sim/scenario/comprehensive_market/agents.py
```diff
@@ -0,0 +1,33 @@
...
+BG_WALLET = WalletConfig("bg", "pin")
+
+# Set up wallets for parties submitting market orders
+MO_WALLETS = (
+    WalletConfig("mo_trader_a", "pass"),
+    WalletConfig("mo_trader_b", "pass"),
+    WalletConfig("mo_trader_c", "pass"),
+    WalletConfig("mo_trader_d", "pass"),
+    WalletConfig("mo_trader_e", "pass"),
+)
```
TomMcL: It would be good if we made the number of these a customisable parameter on the scenario. We can just use a base username then shove a number on the end to generate.

cdummett: Sorry yes was definitely the plan, haven't finished all the changes for this scenario yet.

vega_sim/scenario/common/agents.py
```diff
@@ -889,24 +897,30 @@ def step(self, vega_state: VegaState):
...
 
-    def _submit_order(self):
+    def _submit_order(self, vega_state: VegaState):
 
-        best_bid_price, best_offer_price = self.vega.best_prices(
-            market_id=self.market_id
-        )
+        # Calculate mean_buy_price and mean_sell_price of price distribution
+        if (isinstance(self.spread, type(None))) or (
+            isinstance(self.price_process, type(None))
```
TomMcL: `self.spread is None`

vega_sim/scenario/common/agents.py
```diff
@@ -889,24 +897,30 @@ def step(self, vega_state: VegaState):
...
+            # If agent does not have price_process data, offset orders from best bid/ask
+            best_bid_price, best_offer_price = self.vega.best_prices(
+                market_id=self.market_id
+            )
+            mean_buy_price = best_bid_price
+            mean_sell_price = best_offer_price
+        else:
+            # If agent does have price_process data, offset orders from market price
+            mean_buy_price = self.price_process[self.current_step] - self.spread / 2
+            mean_sell_price = self.price_process[self.current_step] + self.spread / 2
```
TomMcL: I like the logic, but can we rename from `mean` since it isn't really a mean? `buy_price_reference` or something?

vega_sim/scenario/comprehensive_market/agents.py
```diff
@@ -0,0 +1,25 @@
...
+lp_agents = 5
+LP_WALLETS = [WalletConfig("lp_trader_" + str(i), "pass") for i in range(lp_agents)]
+
+# Set up wallets for parties submitting market orders
+mo_agents = 5
+MO_WALLETS = [WalletConfig("mo_trader_" + str(i), "pass") for i in range(mo_agents)]
+
+# Set up wallets for parties submitting / cancelling limit orders
+lo_agents = 20
+LO_WALLETS = [WalletConfig("lo_trader_" + str(i), "pass") for i in range(lo_agents)]
```
TomMcL: Could we make all of these configurable in the scenario itself rather than here? (i.e. as a `num_market_order_agents` parameter)

TomMcL (changes requested):
Mostly looking good 👍 , couple of small comments

TomMcL (approved):
lgtm

PR #136, opened 2022-09-01
feat: Add step to clean intermediate ui dirs to makefile

davidsiska-vega:
Add step to makefile for when console changes so much that you have to do a clean build.


TomMcL (approved):
LGTM

PR #137, opened 2022-09-01
feat: Updating tests and code to remove liquidity provision within market creation proposal

TomMcL:
### Description

Market creation proposals can no longer include a liquidity commitment, so remove that from everywhere (but ensure there is still a commitment made after that where necessary)

### Testing

Updated tests to pass

### Breaking Changes

Externally none, but change in workflow means you'll have to manually submit liquidity after proposing

### Closes

None


davidsiska-vega (approved):
A quick glance says it looks fine.

TomMcL:
Merged, something's up with the poetry requirements change tracking thing but I don't think it's related to this PR

PR #138, opened 2022-09-01
fix: the first liquidity submission isn't an amendment but 1st commit…

davidsiska-vega:
I like to have this example ship-shape ;-).


TomMcL (approved):
LGTM. I tend to have console=False since I don't want to assume people have UI set up but I think if they don't nothing will happen anyway.

No idea what's going on with this requirements check..

PR #139, opened 2022-09-01
feat: Move github actions to python 3.10 and fixing requirements check

TomMcL:
### Description

Updating python in GitHub actions to 3.10 and updating the requirements test as poetry now prints a string about building an environment before outputting the requirements, so we can't blindly compare string outputs.

### Testing

Tests pass

### Breaking Changes

None

### Closes

None


witgaw (approved):
LGTM, obviously I've computed all the hashes locally and checked that they match =]

davidsiska-vega (approved):
If it passes the tests then it must be fine in this case.

PR #141, opened 2022-09-06
test: Try not waiting for auction limit order to stop intermittent test breakages

TomMcL:
### Description

Market Sim is failing intermittently when waiting for this order to submit on Jenkins. This should be a safe non-failing order that we don't need the ID back for so try skipping the waiting and see if this resolves.

### Testing

No additional tests needed. Change covered by existing integration

### Breaking Changes

None

### Closes

None


cdummett (approved):
LGTM

PR #142, opened 2022-09-06
feat: python nb toolkit

cdummett:
### Description
Toolkit for creating Python parameter test notebooks added. DemoNotebook added to demo features.

### Testing
None

### Breaking Changes
None

### Closes
None


parameter_results/toolkit.py
```diff
@@ -0,0 +1,795 @@
...
+
+    def __init__(self, path, dt) -> None:
+        """Inits object by reading data stored in json and csv files in the given path.
+
+        Method loads test settings from the 'run_config.json' and test results from csv
+        files matching the FILE_PATTERN format. Test result data is converted into a
+        list of pandas.DataFrame objects.
+
+        Args:
+            path (str): Path to directory containing parameter test files.
```
TomMcL: Missing docstring for dt

parameter_results/toolkit.py
```diff
@@ -0,0 +1,795 @@
...
+
+        self.data_raw = self._load_data_raw()
+        self.data_avg = self._load_data_avg()
+        self.data_lob = self._load_data_lob()
+
+        self._default_widgets()
+
+    def plot_results(
+        self,
+        variables: list = None,
```
TomMcL: Optional[List[str]]

parameter_results/toolkit.py
```diff
@@ -0,0 +1,795 @@
...
+            for _ in range(len(res_ask)):
+                tmp = res_ask[_].split(": ")
+                if len(tmp) == 2:
+                    result = {}
+                    result["Time Step"] = time_step
+                    result["Price"] = float(tmp[0])
+                    result["Volume"] = float(tmp[1])
+                    result["Side"] = "Ask"
+                    new_data_list.append(result)
+            if time_step != 289:
```
TomMcL: 289?

TomMcL (changes requested):
LGTM, a couple of small tweaks (particularly let's try and get rid of that magic number)

parameter_results/toolkit.py
cdummett: Yep good catch. this is left over from the original notebooks. I will find a way around it now.

TomMcL (approved):
LGTM. Generally I'd say clear the output in a notebook but probably good to have the examples here

cdummett:
Thanks, yeah thats good advice, hadn't thought of that. Will do that for future notebooks 👍

Issue #129, opened 2022-08-24
Create a "momentum" trading agent


davidsiska-vega:
Create a trading agent that follows a momentum trading strategy (which can be configured based on various parameters). 

Step 1 of solving this issue would be finding some basic literature on what momentum strategies look like

TomMcL:
[TA-Lib](https://mrjbq7.github.io/ta-lib/) could be useful for this

PR #143, opened 2022-09-07
feat: Add momentum trader in scenario

Galen-Cao:
### Description

Add agent that can follow multiple momentum strategies to trade: currently, the agent can follow RSI/ STOCHRSI/ APO/ CMO/ MACD strategies and be configured based on some parameters. 

### Testing

Test the performance of momentum agent in notebook, none integration testes added

### Breaking Changes

None 
### Closes

Issue #129


examples/agent_market/MomentumAgent.py
```diff
@@ -0,0 +1,100 @@
...
+                granularity=Granularity.HOUR,
+                start="2022-08-11 12:00:00",
+                end="2022-08-23 21:00:00",
+            ),
+            get_historic_price_series(
+                product_id="ETH-USD",
+                granularity=Granularity.HOUR,
+                start="2022-08-23 22:00:00",
+                end="2022-09-05 09:00:00",
+            ),
```
TomMcL, line 43: I need to fix this up to do this internally 😄

vega_sim/scenario/common/agents.py
```diff
@@ -1129,3 +1138,235 @@ def initialise(self, vega: VegaServiceNull):
...
+            amount=self.initial_asset_mint,
+        )
+
+        self.pdp = self.vega._market_pos_decimals.get(self.market_id, {})
+        self.mdp = self.vega._market_price_decimals.get(self.market_id, {})
+        self.adp = self.vega._asset_decimals.get(self.asset_id, {})
+        self.vega.wait_for_total_catchup()
+
+    def step(self, vega_state: VegaState):
+        self._collect_prices()
```
TomMcL: If this isn't in `Vega_state` we should think about adding then we can just read from there

vega_sim/scenario/common/agents.py
```diff
@@ -1129,3 +1138,235 @@ def initialise(self, vega: VegaServiceNull):
...
+        signal = self.momentum_func_dict.get(self.momentum_strategy[0])()
```
TomMcL: The `get` isn't too useful here since returning `None` would just result in attempting to call `None()` which would break anyway, but can leave this time.

vega_sim/scenario/common/agents.py
```diff
@@ -1129,3 +1138,235 @@ def initialise(self, vega: VegaServiceNull):
...
+                price = (
+                    best_ask + self.offset_levels / 10**self.mdp
+                    if signal == 1
+                    else best_bid - self.offset_levels / 10**self.mdp
+                )
+                self.vega.submit_order(
+                    trading_wallet=self.wallet_name,
+                    market_id=self.market_id,
+                    order_type="TYPE_LIMIT",
+                    time_in_force="TIME_IN_FORCE_IOC",
```
TomMcL: Ah so we're using limit orders as market orders with fixed worst prices rather than staying on the book?

vega_sim/scenario/common/agents.py
```diff
@@ -1129,3 +1138,235 @@ def initialise(self, vega: VegaServiceNull):
...
+            slow_period=self.momentum_strategy[1]["slow_period"],
+            signal_period=self.momentum_strategy[1]["signal_period"],
+        )
+        self.indicators.append(macd - signal)
+
+        if len(self.indicators) == 1:
+            return 0
+
+        if self.indicators[-2] < 0 and self.indicators[-1] >= 0:
+            trade_signal = 1
```
TomMcL: Can we turn trade_signal into an enum please? Will be a little clearer what each number means

vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -208,12 +210,34 @@ def set_up_background_market(
...
+            momentum_strategy=(
+                "RSI",
+                {
+                    "period": 14,
+                },
+            ),
+            send_limit_order=True,
+            offset_levels=20,
+            tag=str(tag),
+        )
```
TomMcL: I wouldn't necessarily add this to the ideal market maker v2 since that's a fairly 'pure' one but maybe let's add to the comprehensive scenario?

TomMcL (changes requested):

TomMcL:
Why are we implementing the indicators here rather than using TALib?

vega_sim/scenario/common/agents.py
Galen-Cao: I have added midprice into Vega_state, so we can read from there now.

vega_sim/scenario/common/agents.py
Galen-Cao: I added `self._RSI`in the get method, so avoid to call `None` if there is no momentum strategy corresponded.

vega_sim/scenario/common/agents.py
Galen-Cao: Have changed the momentum agent to allow it multi types of limit orders, like GTC/GTT, and set the default one to be IOC.

TomMcL (approved):
LGTM

PR #144, opened 2022-09-07
test: Experiment with removing excessive time forwarding

TomMcL:
### Description

Remove forwarding time whilst waiting for proposals to pass. Just do the once. This should make proposals always visible when combines with the physical waits, but will need keeping an eye on.

### Testing

Tests pass

### Breaking Changes

None

### Closes

None


Galen-Cao (approved):
LGTM

PR #146, opened 2022-09-09
feat: Update vega version and move to go 19

TomMcL:
### Description

Moving to Go 1.19 now that core has so that tests pass

### Testing

Integration tests passed locally

### Breaking Changes

Should be none

### Closes

None


Galen-Cao (approved):
LGTM

PR #150, opened 2022-09-13
fix: vega_wallet login method

### LINKED ISSUE #149: vega_wallet methods do not create pub_keys attributes in same convention
## Background:
The `vega_wallet` class uses its `pub_keys` attribute to store public keys for various wallets which can be used by the `submit_transaction` method to sign transactions.

The attribute `pub_keys` is created either when a wallet is created via the `create_wallet` method or when it is logged into via the `login` method.

The submit_transaction method expects a dictionary of wallet names mapped to public keys, e.g:
`self.pub_keys = {'wallet_name_1': "public key", 'wallet_name_2': "public key"}`

## Issue:

- The `create_wallet` method **does** correctly initialise the `pub_keys` attribute.
- The `login` method **does not** correctly initialise the `pub_keys` attribute.

Therefore existing wallets that are "logged into" are currently not compatible with other `vega_wallet` methods (crucially the `submit_transaction` method).

## Todo:
Fix the `login` method so it correctly initialises the `pub_keys` attribute.

cdummett:
### Description
Small change to the `login` method of the `vega_wallet` class.

Change required to ensure method correctly initialises the `pub_keys` attribute in the same format as the `create_wallet` method and the same format expected by the `submit_transaction` method.

### Testing
Initially `test_base_service_wallet_login` in `test_service.py` failing. Response from requests mocker updated to reflect changes. Now passing tests.

### Breaking Changes
Hopefully none.

### Closes
#149

### DIFF
--- a/tests/vega_sim/test_service.py
+++ b/tests/vega_sim/test_service.py
@@ -74,11 +74,8 @@ def test_base_service_wallet_login(stub_service: StubService):
         )
         req_mocker.get(
             WALLET_KEY_URL.format(wallet_server_url="localhost:TEST_WALLET"),
-            json=lambda req, _: {"keys": 'TEST_PHRASE[{"name": "default_key"}]'},
+            json=lambda req, _: {"keys": [{"pub": "TEST_NAME"}]},
         )
         stub_service.login("TEST_NAME", "TEST_PHRASE")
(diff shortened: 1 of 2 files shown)


TomMcL (approved):
LGTM

PR #154, opened 2022-09-20
feat: network service

### LINKED ISSUE #96: Add functionality allowing running of testnet bots on vega-market-sim
We generally want to improve testnet bots. There are two directions to tackle this in:
1) develop good bots in vega-market-sim and then make them run on testnet
2) run testnet bots on vega-market-sim and try for some quick wins. 

This issue is about 2). Once we're able to do this people can easily play with testnet bot parameters / logic to see what effect changing them is likely to have before deploying to testnet and disrupting things there.

cdummett:
### Description
Changes to vega_market_sim to allow scenarios to be run on public and internal Vega networks.

- VegaServiceNetwork added
- VegaEnvironment added
- Fairground scenario added
- modifications made to agents and scenario runners

### Testing
None

### Breaking Changes
None

### Closes
#96


vega_sim/environment/environment.py
```diff
@@ -365,3 +367,83 @@ def step(self, vega: VegaService) -> None:
...
+    def _run(self, vega):
+
+        state_values = []
+
+        # Initialise agents without minting assets
+        for agent in self.agents:
+            agent.initialise(vega=vega)
+
+        for i in range(self.n_steps):
+            time.sleep(1)
```
TomMcL: Let's make the sleep time configurable

vega_sim/environment/environment.py
```diff
@@ -365,3 +367,83 @@ def step(self, vega: VegaService) -> None:
...
+
+    def _run(self, vega):
+
+        state_values = []
+
+        # Initialise agents without minting assets
+        for agent in self.agents:
+            agent.initialise(vega=vega)
+
+        for i in range(self.n_steps):
```
TomMcL: Also we should probably have the ability to not care about n_steps and just run forever. Maybe when n_steps == -1, or just as a separate flag

vega_sim/environment/environment.py
```diff
@@ -365,3 +367,83 @@ def step(self, vega: VegaService) -> None:
...
+            time.sleep(1)
+            print(f"Iteration={i}")
```
TomMcL: Drop the print

vega_sim/environment/environment.py
```diff
@@ -365,3 +367,83 @@ def step(self, vega: VegaService) -> None:
...
+            agent.finalise()
+
+        if self._state_extraction_fn is not None:
+            state_values.append(self._state_extraction_fn(vega, self.agents))
+            return state_values
+
+    def step(self, vega: VegaServiceFairground) -> None:
+        t = time.time()
+        state = self.state_func(vega)
+        print(f"   Get state took {time.time() - t} seconds.")
```
TomMcL: Drop the print, but let's keep the logging message as a debug level log

TomMcL: Although I'm concerned if the state function is taking more than a second to generate

vega_sim/environment/environment.py
```diff
@@ -365,3 +367,83 @@ def step(self, vega: VegaService) -> None:
...
+        for agent in (
+            sorted(self.agents, key=lambda _: random.random())
+            if self.random_agent_ordering
+            else self.agents
+        ):
+            t2 = time.time()
+            agent.step(state)
+            print(f"   Agent {agent.wallet_name} step took {time.time() - t2} seconds.")
```
TomMcL: As above

vega_sim/scenario/common/agents.py
```diff
@@ -2,6 +2,7 @@
 from dataclasses import dataclass
 from multiprocessing.connection import wait
 from random import random
+from telnetlib import SE
```
TomMcL: Delete

vega_sim/scenario/common/agents.py
```diff
@@ -905,6 +914,15 @@ def initialise(self, vega: VegaServiceNull):
...
         self.vega.wait_fn(2)
 
+    def initialise_for_fairground(self, vega: VegaServiceFairground):
+        self.vega = vega
+        self.market_id = [
+            m.id
+            for m in self.vega.all_markets()
+            if m.tradable_instrument.instrument.name == self.market_name
+        ][0]
+        self.asset_id = self.vega.find_asset_id(symbol=self.asset_name)
```
TomMcL, line 1408: Why do we need this? Can't we just use the mint flag like above?

vega_sim/fairground_service.py
```diff
@@ -0,0 +1,67 @@
...
+        "fairground",
+        "--automatic-consent",
+    ]
+
+    process = subprocess.Popen(args=wallet_args)
+
+    return process
+
+
+class VegaServiceFairground(VegaService):
```
TomMcL: As discussed I think the main thing here is let's remove reference to `Fairground` and name everything something more general. `VegaServiceRemote`? `External`? `Network`?

vega_sim/fairground_service.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(
+        self,
+        warn_on_raw_data_access: bool = False,
```
TomMcL: And then instead of the hardcoded strings we can have some `Config` object (thinking data class) with a default one for fairground/eventually main net which people can pass in.

TomMcL (changes requested):

vega_sim/fairground_service.py
cdummett: Sounds good, `Network` makes sense to me and aligns with the terminology in the vega wallet docs.

vega_sim/environment/environment.py
cdummett: Yep, no idea why this is taking so long... even submitting a transaction using the VegaWallet can take 1-2s at the moment. Causes big issues trying to run the BackgroundMarket agent and would probably cause issues with the MarketMaker also.

vega_sim/scenario/common/agents.py
cdummett: Left over from my first attempt at getting agents running on Fairground, have removed now.

cdummett:
Closing this mess. Replaced with PR #172.

PR #155, opened 2022-09-21
feat: Update vega version + remove old genesis flags

TomMcL:
### Description

Removing a couple of genesis keys which have been removed from Vega core

### Testing

Tests now pass

### Breaking Changes

None

### Closes

None


cdummett (approved):
LGTM

PR #156, opened 2022-09-21
chore: update the learning agent so it at least runs

davidsiska-vega:
### Description

The RL agent "trains" and runs again


vega_sim/reinforcement/learning_agent.py
```diff
@@ -208,17 +208,18 @@ def move_to_cpu(self):
...
-        market_name = f"BTC:DAI_{self.tag}"
+        market_name = f"ETH:USD_{self.tag}"
         self.step_num = 0
 
         # Get market id
+        all_markets = self.vega.all_markets()
         self.market_id = [
             m.id
-            for m in self.vega.all_markets()
+            for m in all_markets
```
TomMcL: Why?

vega_sim/reinforcement/learning_agent.py
```diff
@@ -208,17 +208,18 @@ def move_to_cpu(self):
     def initialise(self, vega: VegaServiceNull):
         # Initialise wallet
         super().initialise(vega=vega)
-        market_name = f"BTC:DAI_{self.tag}"
+        market_name = f"ETH:USD_{self.tag}"
```
TomMcL: This should be a flag passed in to the agent these days. Though this isn't making it worse at least :)

vega_sim/reinforcement/market_sim.py
```diff
@@ -5,187 +5,71 @@
...
-    AUCTION1_WALLET,
-    AUCTION2_WALLET,
-    OptimalMarketMaker,
-    MarketOrderTrader,
-    LimitOrderTrader,
-    OpenAuctionPass,
-)
-from .plot import plot_simulation
+
+from plot import plot_simulation
```
TomMcL:
```suggestion
from vega_sim.reinforcement.plot import plot_simulation
```

vega_sim/reinforcement/market_sim.py
```diff
@@ -5,187 +5,71 @@
...
+        block_length_seconds=1,
+        buy_intensity=10,
+        sell_intensity=10,
+        q_upper=50,
+        q_lower=-50,
+        kappa=50,
+        sigma=0.5,
+        num_steps=72,
+        state_extraction_fn=state_fn,
+    )
```
TomMcL: Why all hardcoded now?

vega_sim/reinforcement/test_setup.py
```diff
@@ -0,0 +1,54 @@
...
+from vega_sim.parameter_test.parameter.loggers import (
+    ideal_market_maker_single_data_extraction,
+    target_stake_additional_data,
+    v1_ideal_mm_additional_data,
+    tau_scaling_additional_data,
+    limit_order_book,
+)
+
+
+if __name__ == "__main__":
```
TomMcL: Do we need this module? If so, should be renamed from `test_`. But I suspect we can just delete?

vega_sim/reinforcement/market_sim.py
```diff
@@ -5,187 +5,71 @@
...
     WALLET as LEARNING_WALLET,
 )
-from vega_sim.environment.agent import Agent
+
+from vega_sim.scenario.registry import IdealMarketMakerV2
+from vega_sim.parameter_test.parameter.loggers import (
+    ideal_market_maker_single_data_extraction,
+    target_stake_additional_data,
+    v1_ideal_mm_additional_data,
+    tau_scaling_additional_data,
```
TomMcL: Suspect most of these imports are un-needed?

vega_sim/reinforcement/market_sim.py
TomMcL: We may want to make this a more fun environment anyway though later

TomMcL (changes requested):

vega_sim/reinforcement/learning_agent.py
davidsiska-vega: At some point I pulled it out as a named variable for debugging - I'll put it back as it was - agree this change isn't needed.

vega_sim/reinforcement/learning_agent.py
davidsiska-vega: Yes, good point, let me try to do that.

vega_sim/reinforcement/test_setup.py
davidsiska-vega: I actually don't know what this does 🤣 . If you could suggest the change I'd be very much grateful.

vega_sim/reinforcement/market_sim.py
davidsiska-vega: Yes, I can clean this up.

vega_sim/reinforcement/market_sim.py
davidsiska-vega: Removed hardcoded.

vega_sim/reinforcement/market_sim.py
davidsiska-vega: Removed

vega_sim/reinforcement/learning_agent.py
davidsiska-vega: Done

vega_sim/reinforcement/test_setup.py
TomMcL: The `if` block? It ensures that the code is only run if this is the module which is explicitly being 'called' with `python -m vega_sim.reinforcement.test_setup` etc. Otherwise it would be executed whenever it is imported by another module

TomMcL: Python gives a `__name__` value to each module which you can use for things like generating loggers named for each file (with the standard `logger = logging.getLogger(__name__)`). For the specific file which it's running it gives the name `__main__`.

davidsiska-vega: Excellent, thanks for explaining.

TomMcL (approved):
LGTM 
I think we may want to change the learning agent so that the 'learning' is fully contained within the `Agent` (e.g. we call something like `agent.learn()` after each step), meaning folks can just implement their agent however they want (or have that `learn` step just `pass` and have a set strategy) and ignore everything outside

PR #162, opened 2022-09-23
feat: Better path calculation for directories

TomMcL:
### Description

Updating path calculation logic to work for python 3.9 and 3.10

### Testing

Locally tested


cdummett (approved):
LGTM

PR #164, opened 2022-09-26
Wallet upgrades

### LINKED ISSUE #163: Upgrade AgentWithWallet class to allow use of public keys
### Description:
Currently agents with wallets are initiated with a wallet name and a wallet pass. It would be useful if wallet command could be parsed a specific public key to use, this argument could either be an index of a key to use or the name of a key (provided in the meta-data field).

This is is particularly useful for running agents on networks as existing wallets have to be used (it might be cleaner to have one wallet per scenario with a PK for each individual agent).

Changes are required to a lot of the code-base to implement this. It is worth ensuring default arguments can be used without having to update every example / scenario.

### Changes Required:
- [ ] Update the `pub_keys` property to always be a list (need to take care with how wallets with a single key are handled).
- [ ] Update the `submit_transaction` method with a `pub_key` arg.
- [ ] Update all functions where the wallet property `pub_keys` is used.
- [ ] Update all functions where the wallet method `submit_transaction` is called.

### Testing Required:
- [ ] Check all tests pass.
- [ ] Check all examples run.
- [ ] Check parameter experiments run.

cdummett:
### Description
The `VegaWallet` class and `SlimWallet` class have been upgraded to allow use of wallets with multiple key-pairs:

- The `public_keys` property now maps wallet names to dictionaries mapping key names to public keys.
- The `public_key` method now returns the public key for the specified `wallet_name` AND `key_name`.
- The `submit_transaction` method has an added argument for the name of the key pair to use.
- All calls to the above methods have been updated to include the new additional `key_name` argument.

### Testing
Passing all tests.

### Breaking Changes
Updating classes changed outputs of class methods. Outputs from request mockers in test files updated to reflect changes.

The `SlimWallet` class currently has no `generate_keypair` method so technically users will not be able to create wallets with multiple key-pairs using the `SlimWallet` class. This is fine for now as this PR is focused on adding functionality to the `VegaWallet` class. Changes to the `SlimWallet` class have only been made to align the two classes.

### Closes
#163


vega_sim/wallet/slim_wallet.py
```diff
@@ -116,13 +118,23 @@ def login(self, name: str, **kwargs) -> None:
...
+    ):
         # if self.remaining_until_height_update <= 0:
         #     self.block_height = self.core_client.LastBlockHeight(
         #         core_proto.LastBlockHeightRequest()
         #     ).height
         #     self.remaining_until_height_update = self.height_update_frequency
 
+        pub_key = self.public_key(name=name, key_name=key_name)
+
+        print(pub_key)
```
TomMcL: Remove the print

vega_sim/wallet/vega_wallet.py
```diff
@@ -115,23 +116,32 @@ def generate_keypair(
...
 
-    def get_keypairs(self, wallet_name: str) -> Dict:
+    def get_keypairs(self, wallet_name: str) -> list:
         headers = {"Authorization": f"Bearer {self.login_tokens[wallet_name]}"}
         response = requests.get(
             WALLET_KEY_URL.format(wallet_server_url=self.wallet_url), headers=headers
         )
         response.raise_for_status()
-        return response.json()["keys"]
+        return {key["name"]: key["pub"] for key in response.json()["keys"]}
```
TomMcL, line 134: This is still a dict, so doesn't need the type change above

TomMcL (changes requested):
Mostly looks reasonable, a couple of small changes.

One thing I think we probably should change is that returning the first key from the dictionary turned into a list feels a little fragile (it could change easily). Could we instead have some 'default' name which could be a fixed global variable then set/read that?

vega_sim/wallet/vega_wallet.py
cdummett: > Mostly looks reasonable, a couple of small changes.
> 
> One thing I think we probably should change is that returning the first key from the dictionary turned into a list feels a little fragile (it could change easily). Could we instead have some 'default' name which could be a fixed global variable then set/read that?

Ah that's a nice idea, I'll use "Key 1" as the default as I think that's the default the `vega wallet create` command creates. That will give us some crossover with existing wallets as well (in most cases if you are using an existing wallet you will have to pass a key name though).

cdummett:
Ah that's a nice idea, I'll use "Key 1" as the default as I think that's the default the `vega wallet create` command creates. That will give us some crossover with existing wallets as well (in most cases if you are using an existing wallet you will have to pass a key name though).

vega_sim/wallet/vega_wallet.py
cdummett: Not sure I understand sorry? This line reformats the dict using a dict comprehension no?

vega_sim/wallet/slim_wallet.py
```diff
@@ -17,6 +18,9 @@
 
 logger = getLogger(__name__)
 
+dotenv.load_dotenv()
+DEFAULT_KEY_NAME = os.environ.get("DEFAULT_KEY_NAME")
```
TomMcL: Can we move this read inside the `__init__` of the wallets? e.g. have a `DEFAULT_KEY_NAME = 'foo'` out here then inside `self.default_key = os.environ.get("DEFAULT_KEY_NAME", DEFAULT_KEY_NAME)`. I like to keep as much as possible out of running at import time (plus reading at class creation time lets you vary it if you want at a later time point)

TomMcL: Let's also rename to something more Vega specific (VEGA_DEFAULT_KEY_NAME would work)

vega_sim/wallet/slim_wallet.py
```diff
@@ -17,6 +18,9 @@
 
 logger = getLogger(__name__)
 
+dotenv.load_dotenv()
```
TomMcL: By a similar token to the below, let's move the dotenv load inside the `__init__`s of the services so we only modify someone's env when the sim is doing something which needs it.

vega_sim/wallet/vega_wallet.py
TomMcL: It reformats the dict indeed but (unless I'm missing something) the function is still returning a different type of dict, not a list?

.env
```diff
@@ -1,2 +1,3 @@
 VEGA_SIM_VEGA_TAG=a7ce1fa22b1bd301753367e4798141c6ffc7ecb0
 VEGA_SIM_CONSOLE_TAG=develop
+DEFAULT_KEY_NAME="Key 1"
```
TomMcL: Newline at end of file

TomMcL (changes requested):
Looking pretty good. A couple of suggestions on structuring which you're welcome to argue against 😄

vega_sim/wallet/vega_wallet.py
cdummett: Ah I am with you now sorry, forgot I changed the typing at the top! I will change that.

vega_sim/wallet/slim_wallet.py
cdummett: No that sounds good, I like that a lot more 👍

TomMcL (approved):
LGTM. Same tests are failing on develop so I think we're good to go

Issue #160, opened 2022-09-23
Add functionality to log every transaction/forwarding sent to null chain and replay it exactly


TomMcL:
There are times when it would be useful to log exactly what transactions have been sent to the core and reply those without the delay of recalculating what they should be. Add functionality to save these and each time the chain has been manually forwarded to enable this.

PR #167, opened 2022-09-27
feat: Store transaction replay

TomMcL:
### Description

Adding functionality to store all transactions/forwardings/faucetings in a given Vega session and allow them to be replayed directly without worrying about what logic generated them

### Testing

Manually tested running. Created an integration test which confirms using a scenario that rerunning produces the same Open Interest at least

### Breaking Changes

None

### Closes

#160


davidsiska-vega (approved):
Pretty awesome. I would suggest that in the test we can also try to match all party balances but I am happy for this to be merged as is and do that as an additional improvement later.

PR #170, opened 2022-09-28
fix: SlimWallet default key name

cdummett:
### Description
One-liner change to fix default key name in the SlimWallet which was causing example run with following command to fail.

`python -m examples.nullchain`

### Testing
Passing all tests and examples.


TomMcL (approved):
LGTM

PR #172, opened 2022-09-28
feat: run agents on vega networks

### LINKED ISSUE #96: Add functionality allowing running of testnet bots on vega-market-sim

cdummett:
### Description
Adds functionality to run market-sim agents on a vega network using an existing wallet.

### Testing
Passing all tests and examples.

To further test; set the following environment variables in  `.env` to a wallet name and wallet passphrase on your local machine. Make sure you have some fDAI asset on the wallet you want to run!

- `VEGA_USER_WALLET_NAME`
- `VEGA_USER_WALLET_PASS`

Then run the following command to run the test scenario.

`python -m vega_sim.scenario.network.scenario --network stagnet3`

You should see market orders and limit orders start to be placed by a `MarketOrderTrader` and `LimitOrderTrader` respectively

### Closes
#96


vega_sim/scenario/network/scenario.py
```diff
@@ -0,0 +1,136 @@
...
+    parser.add_argument("--debug", action="store_true")
+
+    args = parser.parse_args()
+
+    logging.basicConfig(level=logging.INFO if not args.debug else logging.DEBUG)
+
+    scenario = Fairground(
+        num_steps=200,
+        step_length_seconds=1,
+        market_name="UNIDAI Monthly (30 Jun 2022)",
```
TomMcL, line 128: In a future PR we should make this all configurable so someone can set this scenario up to just run this and trigger a bunch of customisable bots

vega_sim/scenario/network/scenario.py
```diff
@@ -0,0 +1,136 @@
...
+        self.market_id = [
+            m.id
+            for m in vega.all_markets()
+            if m.tradable_instrument.instrument.name == self.market_name
+        ][0]
+        market_info = vega.market_info(market_id=self.market_id)
+        self.asset_name = (
+            market_info.tradable_instrument.instrument.future.settlement_asset
+        )
+
```
TomMcL, line 62: I assume currently if one runs this all the traders try to trade with the same key and cause chaos? 😄

TomMcL (approved):
LGTM

vega_sim/scenario/network/scenario.py
cdummett: Sounds good, I'll raise a PR early next week for this. Out of interest do you think scenarios running on Vega networks should be kept separate from scenarios run on nullchain? Or should we modify adhoc.py to be capable of running scenarios on both services?

cdummett: I guess the later would be nice because it would let us test a service on nullchain and then deploy it to a network with no changes.

vega_sim/scenario/network/scenario.py
cdummett: Yep exactly 😆 Now the multi-key wallets PR is merged I'll upgrade this scenario.

PR #177, opened 2022-09-30
Margin pics

witgaw:
Add a script setting up market with conditions needed for margin screenshots included in the documentation.

The whole `visualisations` subfolder is just a suggestion, happy to move it elsewhere. Can also get rid of the readme if it's not useful.

Includes Tom's commit with the fix to the issue I was facing locally yesterday.

### DIFF
--- a/examples/visualisations/margins.py
+++ b/examples/visualisations/margins.py
@@ -1,0 +1,146 @@
+import logging
+from collections import namedtuple
+from math import exp
+
+from vega_sim.null_service import VegaServiceNull
+
+WalletConfig = namedtuple("WalletConfig", ["name", "passphrase"])
+
+MM_WALLET = WalletConfig("mm", "pin")
+MM_WALLET2 = WalletConfig("mm2", "pin2")
+TRADER_WALLET = WalletConfig("trader", "password123")
+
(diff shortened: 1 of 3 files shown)


examples/visualisations/margins.py
```diff
@@ -0,0 +1,146 @@
...
+
+wallets = [MM_WALLET, MM_WALLET2, TRADER_WALLET]
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+
+    with VegaServiceNull(
+        run_with_console=True,
+        start_order_feed=True,
+        retain_log_files=True,
```
davidsiska-vega: If things are working I'd set this to False; it's not just the log files but also vega home so chain, data node... and even the log can be GB - though probably not from this script.

davidsiska-vega (commented):
I've left one comment but I'd wait for @TomMcL to review since it has a change in the `null_sevice.py` ...

examples/visualisations/margins.py
witgaw: Good call, thanks - got rid of these now.

TomMcL (dismissed):
LGTM

TomMcL (approved):

PR #179, opened 2022-09-30
feat: ensure faucet worked

TomMcL:
### Description

Adding checks to the minting/fauceting process to ensure that the null chain has registered it before continuing

### Testing

Tests pass


vega_sim/api/governance.py
```diff
@@ -421,11 +421,19 @@ def _make_and_wait_for_proposal(
...
-        lambda p: _proposal_loader(p, data_client),
-    )
+    # Allow one failure, forward once more
+    try:
+        time_forward_fn()
+        proposal = wait_for_acceptance(
+            proposal.reference,
+            lambda p: _proposal_loader(p, data_client),
+        )
+    except ProposalNotAcceptedError:
```
davidsiska-vega, line 431: When I was a lad I was told in school that using exceptions for "normal flow control" is bad practice... but I don't see why it would really be a problem here - we never expect this to be "fast".

davidsiska-vega (approved):

vega_sim/api/governance.py
TomMcL: Really this is just a glorified 'try twice then fail' loop which seemed slightly clearer than a tiny range 😄

PR #181, opened 2022-09-30
fix: average entry price on positions is in market quote decimals

davidsiska-vega:
### Description
Average entry price on positions is in market quote decimals, fixed that.

### Testing
I am feeling lucky.


davidsiska-vega:
Must have been resolved by now, closing.

PR #182, opened 2022-09-30
feat: fairground historic trading scenario

### LINKED ISSUE #178: Create scenario to run agents on a Fairground market
### Description

Recent PRs have added the capability to run `vega-market-sim` agents on any Vega network. This ticket is for creating a scenario which can be deployed by DevOps on a market setup specifically for running `vega-market-sim` agents on. The scenario requires the following features:

1. `IdealMarketMaker` agent (provides a market)
2. `SimpleLiquidityProvider` agents (provide enough liquidity for the expected open  interest of the traders)
3. `MomentumTrader` agents (possibly multiples for each strategy with a variety of strategy parameters).
4. `BackgroundMarket` which follows a historic price-process (ideally a live price-process but historic for now).
5. `MarketOrderAgents` which guarantee trades each step to ensure candles are created for the `MomentumTraders`

Some of the above agents will need some modifications to be able to run with multi-key wallets.

### Extras

As an additional nice to have it should be possible to run the same scenario on a nullchain network. This would allow testing of network parameters which can be given to DevOps when setting up the market. For this the scenario would require the following extra features:

1. `MarketManager` agent which can propose and settle a market (should be configurable so as to not run on fairground).
2. `adhoc.py` modifications to allow network selection

Some more mods might need to be made to the `SlimWallet` to allow multi-key wallets. Original mods were made with the intention the `SlimWallet` would only ever single-key wallets.

cdummett:
### Description
Added a scenario which will eventually be deployed to fairground. Initially scenario will be tested on a nullchain network before being deployed.

Included in this PR are the following updates to make scenario compatible with multi-key wallets.
- Updates to various agents
- Updates to `VegaService` methods
- Updates to `governance` functions

### Using

To run the scenario run the following command:
`python -m vega_sim.scenario.adhoc -s fairground --console --pause`

To run the scenario on a network run the following command (with the necessary wallets setup).
`python -m vega_sim.scenario.adhoc -s fairground --console --pause --network fairground`

### Testing
Passing all tests locally and examples working.

### Closes
Closes #178


vega_sim/scenario/adhoc.py
```diff
@@ -14,6 +16,7 @@ def main():
     parser.add_argument("--graphql", action="store_true")
     parser.add_argument("--pause", action="store_true")
     parser.add_argument("--store", action="store_true")
+    parser.add_argument("--network")
```
TomMcL: Can we add `default` and `choices` arguments to this? Then we can drop the `is None` below and be more explicit about what networks we can handle

vega_sim/scenario/common/agents.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from curses import keyname
```
TomMcL: Another rogue import 😄

vega_sim/scenario/adhoc.py
cdummett: List of choices is a great idea I've made that change.

If the default argument is set in `adhoc.py` and the `is None` logic is removed throughout the codebase than the `run_iteration` method in each scenario really needs to be updated to include the `network` argument (even though it doesn't strictly use it).

To get round this we could set the default `network` value in the parent `Scenario` class too `nullchain`? Means the  default `network` value is being set in two different places though.

cdummett: Ignore the above, I think your way is cleaner after using it for a bit 😆 

I am making `network` a required argument in `run_iteration` and updating all the necessary calls now.

vega_sim/scenario/adhoc.py
```diff
@@ -30,17 +37,29 @@ def main():
...
-        launch_graphql=args.graphql,
-        warn_on_raw_data_access=False,
-        seconds_per_block=scenario.block_length_seconds,
-        transactions_per_block=100,
-        retain_log_files=True,
-        use_full_vega_wallet=False,
-        store_transactions=args.store,
-    ) as vega:
-        scenario.run_iteration(vega=vega, pause_at_completion=args.pause)
+    if args.network == "nullchain":
```
TomMcL: Assuming the list of network options is reasonably fixed I think we'd be best off making this an enum somewhere rather than passing a string around

vega_sim/scenario/curve_market_maker/scenario.py
```diff
@@ -263,6 +263,7 @@ def set_up_background_market(
     def run_iteration(
         self,
         vega: VegaServiceNull,
+        network: str,
```
TomMcL: Can we make these Optional everywhere too please? Given that I think it's totally unused in the default Nullchain case?

TomMcL (changes requested):

vega_sim/scenario/fairground/scenario.py
```diff
@@ -0,0 +1,433 @@
...
+MOMENTUM_MARKET_ORDER_ARGS = {
+    "initial_asset_mint": 1e06,
+    "momentum_strategies": ["MACD", "APO", "RSI", "STOCHRSI", "CPO"],
+    "order_intensity": 5,
+}
+SENSITIVE_MARKET_ORDER_ARGS = {
+    "initial_asset_mint": 1e06,
+    "buy_intensity": 5,
+    "sell_intensity": 5,
+}
```
TomMcL, line 126: It would be nice to move these out to a JSON config file so someone can tweak and pass in without having to edit the code itself, but not in this PR :)

vega_sim/scenario/fairground/scenario.py
```diff
@@ -0,0 +1,433 @@
...
+
+class Fairground(Scenario):
+    """Scenario for simulating a realistic market on Fairground.
+
+    Creates a scenario which can be run locally on a nullchain network and
+    run on an existing public or internal Vega network. The scenario contains
+    the following features:
+
+    • ExponentialShapedMarketMaker agent which provides a market for traders
+    • OpenAuctionPass agents which provide a constant best bid and ask
```
TomMcL: OpenAuctionPass should just be providing a single limit order to help the market exit opening auction, rather than anything constant

vega_sim/scenario/fairground/scenario.py
```diff
@@ -0,0 +1,433 @@
...
+    • MarketOrderTrader agents which provide trades.
+    • MomentumTrader agents which provide trades.
+    • PriceSensitiveMarketOrder agents which provide trades.
+
+    • MarketManger agent which proposes and settles a market (only required
```
TomMcL:
```suggestion
    • MarketManager agent which proposes and settles a market (only required
```

TomMcL (commented):
Also not in this PR, but I'm thinking with the Wallet key_name functionality in now we could probably move away from passing the wallet name in to service functions and *just* pass in the key (after storing the wallet name somewhere and reading it back each time)

vega_sim/scenario/curve_market_maker/scenario.py
cdummett: Sure thing, I will make it optional in any scenario where it is unused, but will keep as needed for scenarios which can run either the `VegaServiceNullchain` service or `VegaServiceNetwork`

vega_sim/scenario/fairground/scenario.py
cdummett: Yeah I agree, I think it would be really nice if we didn't have to pass a huge number of arguments to a scenario but could instead pass in a single argument for which config to use for each agent type.

A bit like in the vega feature tests where you have configs you can use to create a market quickly.

vega_sim/scenario/comprehensive_market/scenario.py
```diff
@@ -321,6 +321,7 @@ def set_up_background_market(
     def run_iteration(
         self,
         vega: VegaServiceNull,
+        network: Optional[str] = None,
```
TomMcL: These should be `Optional[Network]`s right?

vega_sim/scenario/constants.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum, auto
+
+
+class Network(Enum):
+    NULLCHAIN = "nullchain"
+    STAGNET1 = "vegawallet-stagnet1"
```
TomMcL, line 6: Why isn't this just "stagnet1"?

vega_sim/scenario/curve_market_maker/scenario.py
```diff
@@ -321,5 +322,6 @@ def run_iteration(
     ) as vega:
         scenario.run_iteration(
             vega=vega,
+            network="nullchain",
```
TomMcL: Update to enum

vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -283,5 +284,6 @@ def run_iteration(
     ) as vega:
         scenario.run_iteration(
             vega=vega,
+            network="nullchain",
```
TomMcL: enum

vega_sim/scenario/market_crash/scenario.py
```diff
@@ -278,4 +279,4 @@ def run_iteration(
         run_with_console=True,
         seconds_per_block=40,  # Heuristic
     ) as vega:
-        scenario.run_iteration(vega=vega, pause_at_completion=True)
+        scenario.run_iteration(vega=vega, network="nullchain", pause_at_completion=True)
```
TomMcL: enum

vega_sim/scenario/network/scenario.py
```diff
@@ -133,4 +134,4 @@ def run_iteration(
         automatic_consent=True,
         no_version_check=True,
     ) as vega:
-        scenario.run_iteration(vega=vega)
+        scenario.run_iteration(vega=vega, network="nullchain")
```
TomMcL: enum

TomMcL (commented):
Couple of places left to update the str to the enum but looking good other than that

vega_sim/scenario/constants.py
cdummett: Ah yes so this is annoying, the config file for `stagnet1` is called `vegawallet-stagnet1.toml`, I would like to neaten the way VegaServiceNetwork reads config files in the future to make it a bit better and probably put this Enum somewhere else but for now it should be okay

vega_sim/scenario/comprehensive_market/scenario.py
cdummett: Thanks yep have changed throughout

TomMcL (approved):
LGTM

PR #185, opened 2022-10-02
Another RL round

davidsiska-vega:
### Description
Added plotting of learning progress and final PnL.
Added a "heuristic" informed trader strategy to see if we can at least learn the Q-function.
Added another RL agent who doesn't try to optimise the volume; always does MOs of the smallest size.
Added slides with basic theory - it's a question whether to actually merge those - but for now it's a convenient place. 

### Testing
Run the agents; they seem to be doing something if not necessarily learning to make a profit. 

### Breaking Changes
None I hope. Though the PR does touch some of the environment and other agent files.


vega_sim/scenario/common/agents.py
```diff
@@ -1209,7 +1210,7 @@ def __init__(
                 q_lower=self.q_lower,
                 mdp=self.mdp,
                 kappa=self.market_kappa,
-                lmbda=self.market_order_arrival_rate,
+                Lambda=self.market_order_arrival_rate,
```
TomMcL, line 1217: D'oh, meant to change these both to lmbda. We can stick to Lambda for now though :+1:

vega_sim/scenario/common/agents.py
```diff
@@ -1581,22 +1582,30 @@ def initialise(
         self.vega.wait_for_total_catchup()
 
     def step(self, vega_state: VegaState):
+        market_info = self.vega.market_info(market_id=self.market_id)
+        market_in_auction = (
+            not market_info.trading_mode
+            == markets_protos.Market.TradingMode.TRADING_MODE_CONTINUOUS
+        )
```
TomMcL, line 1593: This should be contained within the vega_state object (though it needs updating to be clearer about what is inside)

vega_sim/scenario/common/agents.py
```diff
@@ -1581,22 +1582,30 @@ def initialise(
...
+                self.vega.submit_market_order(
+                    trading_wallet=self.wallet_name,
+                    market_id=self.market_id,
+                    side=trade_side,
+                    volume=np.abs(current_position),
+                    wait=True,
+                    fill_or_kill=False,
+                )
+            except OrderRejectedError:
+                print("Order rejected.")
```
TomMcL: We should use a logger here

vega_sim/scenario/ideal_market_maker_v2/scenario.py
```diff
@@ -66,6 +67,7 @@ def __init__(
         proportion_taken: float = 0.8,
     ):
         self.num_steps = num_steps
+        self.random_agent_ordering = (random_agent_ordering,)
```
TomMcL: Why a tuple?

TomMcL (changes requested):

vega_sim/scenario/ideal_market_maker_v2/scenario.py
davidsiska-vega: This looks like a fat finger at best. Will look.

vega_sim/scenario/common/agents.py
davidsiska-vega: Yes, debug level?

vega_sim/scenario/common/agents.py
davidsiska-vega: Could you offer the change here, I am not sure what you mean exactly.

vega_sim/scenario/common/agents.py
davidsiska-vega: Let me do the change to `lmbda` since there are other changes and I'll need to run the tests again ;-).

vega_sim/scenario/common/agents.py
TomMcL: Yeah I think so, it's an error which you may or may not care about so doing at exception level might well feel noisy. Could do at `warn` but I never much like `warn`

vega_sim/scenario/common/agents.py
davidsiska-vega: Actually, I've changed my mind - I'll do a separate PR - otherwise this will touch even more files / places.

vega_sim/scenario/common/agents.py
```diff
@@ -1581,22 +1582,30 @@ def initialise(
         self.vega.wait_for_total_catchup()
 
     def step(self, vega_state: VegaState):
+        market_info = self.vega.market_info(market_id=self.market_id)
```
TomMcL:
```suggestion
        trading_mode = vega_state.market_state[self.market_id].trading_mode
```

vega_sim/scenario/common/agents.py
TomMcL: Added proposed change, but requires some other line tidying too

davidsiska-vega: Thx!

TomMcL (approved):
LGTM

PR #186, opened 2022-10-03
feat: make multi-key wallets compatible with nullchain

cdummett:
### Description
PR makes changes to wallets, service methods, and agents to make multi-key agents compatible with nullchain simulations. This PR is part of the work to allow a scenario to be run as a simulation on nullchain and then an existing network with no changes to the scenario.

### Testing
Passing all examples and tests.

### Closes
None

### DIFF
--- a/vega_sim/environment/agent.py
+++ b/vega_sim/environment/agent.py
@@ -42,6 +42,9 @@ class AgentWithWallet(Agent):
                 str, The name to use for this agent's wallet
             wallet_pass:
                 str, The password which this agent uses to log in to the wallet
+            key_name:
+                str, optional, Name of key in wallet for agent to use. Defaults
+                to value in the environment variable "VEGA_DEFAULT_KEY_NAME".
         """
         super().__init__()
         self.wallet_name = wallet_name
--- a/vega_sim/environment/agent.py
+++ b/vega_sim/environment/agent.py
@@ -1,6 +1,6 @@
 from abc import ABC
 from dataclasses import dataclass
-from typing import Any, Tuple, Dict
+from typing import Any, Tuple, Dict, Optional
 
 from vega_sim.service import VegaService
 
@@ -27,6 +27,7 @@ class AgentWithWallet(Agent):
         self,
         wallet_name: str,
         wallet_pass: str,
+        key_name: Optional[str] = None,
     ):
         """Agent for use in environments as specified in environment.py.
         To extend, the crucial function to implement is the step function which will
@@ -45,6 +46,7 @@ class AgentWithWallet(Agent):
         super().__init__()
         self.wallet_name = wallet_name
         self.wallet_pass = wallet_pass
+        self.key_name = key_name
 
     def step(self, vega: VegaService):
         pass
@@ -52,7 +54,11 @@ class AgentWithWallet(Agent):
     def initialise(self, vega: VegaService, create_wallet: bool = True):
         super().initialise(vega=vega)
         if create_wallet:
-            self.vega.create_wallet(name=self.wallet_name, passphrase=self.wallet_pass)
+            self.vega.create_wallet(
+                name=self.wallet_name,
+                passphrase=self.wallet_pass,
+                key_name=self.key_name,
+            )
         else:
             self.vega.login(name=self.wallet_name, passphrase=self.wallet_pass)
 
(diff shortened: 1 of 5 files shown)


vega_sim/wallet/slim_wallet.py
```diff
@@ -115,28 +115,37 @@ def _next_nonce(self) -> int:
             self._create_nonces()
         return self.nonces[self.nonce_idx]
 
-    def create_wallet(self, name: str, **kwargs) -> None:
+    def create_wallet(self, name: str, passphrase: str, key_name: str = None) -> None:
```
TomMcL: Can we keep passphrase as explicitly optional here? (And make them both `Optional`)

vega_sim/wallet/vega_wallet.py
```diff
@@ -1,3 +1,4 @@
+from importlib.metadata import metadata
```
TomMcL: erroneous?

vega_sim/wallet/vega_wallet.py
```diff
@@ -31,11 +32,7 @@ def __init__(self, wallet_url: str):
             "VEGA_DEFAULT_KEY_NAME", VEGA_DEFAULT_KEY_NAME
         )
 
-    def create_wallet(
-        self,
-        name: str,
-        passphrase: str,
-    ) -> str:
+    def create_wallet(self, name: str, passphrase: str, key_name: str = None) -> None:
```
TomMcL: `Optional` on the key_name

vega_sim/wallet/vega_wallet.py
```diff
@@ -123,7 +126,6 @@ def generate_keypair(
             json=req,
         )
         response.raise_for_status()
-        return list(response.json()["key"]["pub"])
```
TomMcL: Why stop returning the key?

vega_sim/environment/agent.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         self,
         wallet_name: str,
         wallet_pass: str,
+        key_name: Optional[str] = None,
```
TomMcL, line 30: Can we add a short bit to the docstring too?

vega_sim/service.py
```diff
@@ -268,7 +268,9 @@ def login(self, name: str, passphrase: str) -> str:
         """
         return self.wallet.login(name=name, passphrase=passphrase)
 
-    def create_wallet(self, name: str, passphrase: str) -> str:
+    def create_wallet(
+        self, name: str, passphrase: str, key_name: Optional[str] = None
+    ) -> str:
```
TomMcL, line 273: Can we add a short bit to the docstring too?

TomMcL (changes requested):

vega_sim/wallet/vega_wallet.py
cdummett: Yep thanks, not sure how I keep adding in these weird unused imports...!

vega_sim/wallet/vega_wallet.py
cdummett: Just the output didn't seem to be used in any of the calls in `VegaWallet` or `SlimWallet`. Tbh I didn't check through the entire codebase so have added back in just in case!

vega_sim/environment/agent.py
cdummett: Have added:
```
key_name:
    optional, Name of key in wallet for agent to use. Defaults to
    the value in the environment variable "VEGA_DEFAULT_KEY_NAME".
```
Do you think adding the description of what it defaults too is confusing given it doesn't strictly default to that value here but instead when the `Wallet` class doesn't receive a key name?

vega_sim/service.py
cdummett: See comment above ^^^

vega_sim/environment/agent.py
TomMcL: I think this seems reasonable. Letting people know the 'ultimate' default rather than necessarily what is immediately used seems fine to me

vega_sim/wallet/vega_wallet.py
TomMcL: It's vscode being 'helpful' automatically adding imports, which is useful most of the time until you have a slight typo and press enter 😄

TomMcL (approved):
LGTM

PR #188, opened 2022-10-03
feat: add max margin and LP return

Jiajia-Cui:


vega_sim/parameter_test/parameter/loggers.py
```diff
@@ -126,17 +133,38 @@ def _ideal_market_maker_single_data_extraction(
...
 
     additional_fns = additional_data_fns if additional_data_fns is not None else []
+
+    if len(state_values) == 0:
+        max_margin = margin_lp
+    else:
+        old_logs = state_values[mm_agent.current_step-2]
+        max_margin = old_logs["LP: Max Margin"]
+        max_margin = max(margin_lp, max_margin)
+
```
Galen-Cao, line 151: Could we maybe put calculation of max_margin & max_lockedCapital together to simplify the code? And maybe let `old_logs = state_values[-1]` should have the same result.

TomMcL:
What are the changes in the DemoNotebook here?

vega_sim/parameter_test/parameter/loggers.py
TomMcL: Agreed on both

vega_sim/parameter_test/parameter/loggers.py
```diff
@@ -126,17 +133,38 @@ def _ideal_market_maker_single_data_extraction(
...
+    if len(state_values) == 0:
+        max_lockedCapital = margin_lp + bond_lp
+    else:
+        old_logs = state_values[mm_agent.current_step-2]
+        max_lockedCapital = old_logs["LP: Max Locked Capital"]
+        max_lockedCapital = max(margin_lp + bond_lp, max_lockedCapital)
```
TomMcL: `max_locked_capital`

Jiajia-Cui:
> 

added the new outputs in `experiment.plot_results`

vega_sim/parameter_test/parameter/loggers.py
```diff
@@ -62,6 +66,7 @@ def ideal_market_maker_single_data_extraction(
 def _ideal_market_maker_single_data_extraction(
     vega: VegaServiceNull,
     agents: List[Agent],
+    state_values: List = None,
```
TomMcL, line 73: Still not really sure I like this vs just calculating derived values at a later time, though in the interests of expediency we can probably stick with it for now

TomMcL (changes requested):
Needs a `black` run too :+1:

TomMcL:
Needs another `make black` please 😄 (You can enable in VSCode so that it runs on save, which makes things easier!) 

Also, the DemoNotebook.ipynb file isn't for actual investigation, should just be a useful template to copy into the specific folders for each investigation. This one still looks reasonable though so will leave it for now!

TomMcL (approved):

PR #205, opened 2022-10-11
Feat/ux examples

### LINKED ISSUE #198: Create a scenario where a party doesn't meet margin balance requirement and gets closed out
Create a scenario where a party doesn't meet margin balance requirement and gets closed out:
- case 1: their margin balance is sufficient for the close-out trade and so insurance pool is funded in the process
- case 2: their margin balance is insufficient for the close-out and so balance is taken out of insurance pool to fund them 

I think they could go into `examples/visualisations`. 

I think we should add integration tests to make sure they run even as we change stuff (so that others can rely on them). 

@cdummett there should be feature tests that create these situations though if they use the "simple risk model" (which is only for feature tests) then it will need a bit of work to cook up the cases.

### LINKED ISSUE #199: Create a scenario where a party suffers from loss socialisation
Create a scenario where a party suffers from loss socialisation:
1) Bob opened a position and the market moved in their favour.
2) Alice's position no longer meets margin requirement; her position gets closed out; there is insufficient amount in her margin+insurance pool to cover the shortfall.
3) Bob gets a positive mark-to-market cash flow but not the full amount he'd expect from `position size x price change` due to loss socialisation. 

Do after #198 . 

Again, there should be feature tests demo-ing this.

cdummett:
### Description

Visualisation examples added to allow UX to test console in the following states:

- Party closed out (loss-socialisation not required)
- Party closed out (loss-socialisation required)

PR also adds utility functions to allow visualisation examples to be made faster (doc-strings required).

### Testing

Passing all integration tests.

To manually test, ensure you have the most recent UI with `make ui` and then run the following commands:

`python -m examples.visualisations.closed_out --console --pause`

`python -m examples.visualisations.loss_socialisation --console --pause`

### Closes
closes #198 
closes #199

### DIFF
--- a/examples/visualisations/closed_out.py
+++ b/examples/visualisations/closed_out.py
@@ -1,0 +1,138 @@
+"""closed_out.py
+
+Script launches an example of a market moving against a party with a short-position
+whom is eventually closed out. When the party is closed out they have sufficient
+collateral in their margin account to cover their losses and loss-socialisation is not
+required.
+
+When running the example with the --console flag, the user will be able to inspect the
+positions of the following parties:
+
+    • TRADER_A: party with long position
+    • TRADER_B: party with short position (closed out)
+
+Usage:
+
+    python -m examples.visualisations.closed_out --console --pause
+
+Flags:
+
+    --console   will launch the console
+    --debug     will log debug messages 
+    --pause     will pause the script after each price movement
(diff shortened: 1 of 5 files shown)


Jiajia-Cui (approved):
LGTM

PR #216, opened 2022-10-21
Update env

davidsiska-vega:
### Description
Updated to latest core and console tags

### Testing
run `make test_integration` and manually verified console builds / loads. 

### Breaking Changes
None AFAIK.


cdummett (approved):
Tested with examples launching console - LGTM

PR #221, opened 2022-10-27
feat: flexible market id finder

### LINKED ISSUE #212: Overhaul how agents obtain a market_id
### Description
At the moment a `market_name` is passed to agents who then match it against all markets with a matching tradable instrument name on the network. The `market_id` is then set to the first `market_id` in the list of matches. This method is required for running simulations with the `VegaServiceNull` service as the `market_id` is not known before the simulation.

This method fails when a network has multiple markets using the same tradable instrument name matching `market_name`. This scenario is most likely to occur if a market has been terminated and re-proposed.

### Changes

1. Simplest option is to add an optional `market_id` arg to all agents. This would remove ambiguity when running agents on an existing network (useful for DevOps & Incentives). This would not work for perpetual market-sim markets.
2. To fix the above point agents should be able to search only for active markets and raise an error if there are multiple active markets with a tradable instrument name matching the `market_name`.

cdummett:
### Description
Adds a `find_market_id` method to the vega `Service` class and updates all agents to use it.

Method returns the market id of the market matching the specified market name providing it is in one of the following states:

- `STATE_PENDING`
- `STATE_ACTIVE`
- `STATE_SUSPENDED`

If more than one market with a matching market name is found the method returns the id of the market with the most recently pending timestamp.

PR required to run market-sim bots on networks. Further matching criteria can be added in the future to further improve method.

### Testing
Passing all tests locally.

### Closes
Closes #212

### DIFF
--- a/examples/agent_market/agents.py
+++ b/examples/agent_market/agents.py
@@ -159,7 +159,7 @@ class MarketMaker(StateAgentWithWallet):
             termination_wallet=self.terminate_wallet_name,
             liquidity_commitment=liquidity_commitment,
         )
-        self.vega.forward("2s")
+        self.vega.wait_for_total_catchup()
 
         self.market_id = self.vega.all_markets()[0].id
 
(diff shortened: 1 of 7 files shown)


Jiajia-Cui (approved):
LGTM

PR #227, opened 2022-11-08
feat: add parameter QuantumMultiple

Jiajia-Cui:


vega_sim/api/market.py
```diff
@@ -15,6 +15,7 @@
 • liquidity_monitoring_parameters.auction_extension
```
cdummett: As this is a network parameter experiment, we can remove all these changes to `market.py`

vega_sim/parameter_test/parameter/configs.py
```diff
@@ -261,15 +261,34 @@
 )
 
 MARKET_PARAMETER_DEMO = SingleParameterExperiment(
```
cdummett: Instead of replacing `MARKET_PARAMETER_DEMO` could we add a new experiment (e.g. `MIN_LP_STAKE_QUANTUM`).

cdummett (commented):
Is there a notebook for this experiment as well by the way?

cdummett (approved):

PR #235, opened 2022-11-11
feat: implement batch-market-instructions

### LINKED ISSUE #148: Move any agents who do multiple amends/creates in a step to using batch instructions
Batch instructions have now been [released to Vega core](/var/folders/yj/cjhtlxn90wldd1hvw5lkxnrc0000gn/T/tmpv97jnw6r). Update any agents who can be to use these new order types.

cdummett:
### Description
Feature adds in capability for agents to submit batch-market-instructions.

Agents must build list of `OrderCancellation`, `OrderAmendment`, and `OrderSubmission` proto objects and pass them into the service method `submit_market_instructions`.

The method is setup to allow batches larger than the spam protection network parameter `spam.protection.max.batchSize`, in this case the method splits batch of instructions into smaller batches.

### Testing
Passing all tests locally.

Try running:

`$ python -m vega_sim.scenario.adhoc -s configurable_market --console --pause --debug`

### Closes
Closes #148


vega_sim/api/trading.py
```diff
@@ -408,3 +413,235 @@ def submit_liquidity(
         key_name=key_name,
     )
     logger.debug(f"Submitted liquidity on market {market_id}")
+
+
+def pegged_order(
+    reference: vega_protos.vega.PeggedReference,
+    offset: str,
+) -> vega_protos.vega.PeggedOrder:
```
TomMcL, line 393: I'm not entirely certain this simplifies anything vs just using the proto directly, but it's here now so we can keep it 😄

vega_sim/api/trading.py
```diff
@@ -408,3 +413,235 @@ def submit_liquidity(
...
+    time_in_force = (
+        time_in_force
+        if time_in_force is not None
+        else vega_protos.vega.Order.TimeInForce.TIME_IN_FORCE_UNSPECIFIED
+    )
+
+    time_in_force = get_enum(time_in_force, vega_protos.vega.Order.TimeInForce)
+
+    if pegged_reference is not None:
+        pegged_reference = get_enum(time_in_force, vega_protos.vega.PeggedReference)
```
TomMcL: shouldn't be `time_in_force` here

vega_sim/service.py
```diff
@@ -1528,3 +1533,288 @@ def get_trades(
...
+            elif isinstance(instruction, OrderSubmission):
+                batch_of_submissions.append(instruction)
+            else:
+                raise ValueError(f"Invalid instruction type {type(instruction)}.")
+
+            if (
+                len(batch_of_cancellations)
+                + len(batch_of_amendments)
+                + len(batch_of_submissions)
+                == max_batch_size
```
TomMcL: I don't think it's an issue here, but just as a point of safety on a check like this which is incrementing up I like to use `>=` rather than `==` just to avoid ending up in an infinite loop or similar when you accidentally add two things in one loop :)

TomMcL: Also, given that we can always either increment total length by 1 or raise a `ValueError` maybe it would be fastest here to just use a counter which increments by 1 and is reset alongside the lists (though that's almost certainly a premature optimisation)

vega_sim/api/trading.py
```diff
@@ -408,3 +413,235 @@ def submit_liquidity(
...
+        ("price", price),
+        ("expires_at", expires_at),
+        ("reference", reference),
+        ("pegged_order", pegged_order),
+    ]:
+        if val is not None:
+            setattr(command, attr, val)
+
+    # Return the created and updated OrderSubmission object
+    return command
```
TomMcL, line 581: Can we make the existing amend/cancel/submit functions use these too so that we use the same logic in both paths?

vega_sim/service.py
```diff
@@ -1528,3 +1533,288 @@ def get_trades(
...
+        )
+
+    def submit_instructions(
+        self,
+        wallet_name: str,
+        key_name: Optional[str] = None,
+        cancellations: Optional[List[OrderCancellation]] = None,
+        amendments: Optional[List[OrderAmendment]] = None,
+        submissions: Optional[List[OrderSubmission]] = None,
+    ):
```
TomMcL, line 1745: It feels like it would be nice to also have an interface into this method where one doesn't have to first separately create the individual objects, but I also can't think how to do that without becoming a horrible mess, so I'm open to suggestions and this one seems good too 😄

vega_sim/api/data.py
```diff
@@ -827,3 +827,24 @@ def get_trades(
             )
         )
     return res_trades
+
+
+def get_network_parameter(
+    data_client: vac.VegaTradingDataClientV2,
+    key: str,
+) -> Any:
```
TomMcL: Does this not always return a string? (Or at worst a string/float/int?)

TomMcL: I'm also on the fence on whether this should be in `data_raw` (because we aren't processing the output at all) or `data` (because we're unlikely to want to)

TomMcL: It also might be cool to have a `to_type` (or similar name) arg which takes a function and automatically converts the result to that (so, you don't have to do it externally) e.g. `foo = get_network_parameter(key='bar', to_type=int)`. It'll make the type signature marginally funkier but that's a challenge to the reader

davidsiska-vega (commented):
> The method is setup to allow batches larger than the spam protection network parameter 

I am not sure this is all that useful. Because we have cancellations, then amends then posts the behaviour will be different between your batch fitting in in one go or being split across multiple orders.

cdummett:
> > The method is setup to allow batches larger than the spam protection network parameter
> 
> I am not sure this is all that useful. Because we have cancellations, then amends then posts the behaviour will be different between your batch fitting in in one go or being split across multiple orders.

The thinking was this was needed to allow the market-marker agents to maintain a number of levels on each side of the book greater than the spam protection parameter. Without it, batches of instructions are rejected.

Do you think one of these other options is preferable @davidsiska-vega ?

1. Cap the number of market-maker levels at `spam.protection.max.batchSize`
2. Move the logic for splitting a large batch into multiple batches into the market-maker agents themselves.

vega_sim/service.py
cdummett: Second option sounds great thanks, I'll definitely add that in

vega_sim/api/trading.py
cdummett: Yeah definitely, I was going to do this as a separate PR once this logic was approved, but happy to include in this one

vega_sim/service.py
cdummett: Hm yeah I completely agree I just couldn't think of an elegant way around it either! The way the MMs are setup now it just seemed to make sense to create the objects as you pass through the loop in `_move_side`.

We could definitely add some simpler methods for other agents to use; e.g. batch_submit_orders (pass a list of prices and volumes to submit limit orders at), batch_cancel_orders (pass a list of order ids to cancel)?

vega_sim/api/data.py
cdummett: > Does this not always return a string? (Or at worst a string/float/int?)

Not sure which, but one of those is definitely right! I'll have a look and then change the typing 👍 

> It also might be cool to have a to_type (or similar name) arg which takes a function and automatically converts the result to that (so, you don't have to do it externally) e.g. foo = get_network_parameter(key='bar', to_type=int). It'll make the type signature marginally funkier but that's a challenge to the reader


I like that `to_type` idea as well, I'll add that in. When I do I guess we can treat that as processing the output in service and move the actual `get_network_parameter` method to `data_raw`?

vega_sim/api/trading.py
cdummett: The logic was if data-node or protos are updated in the future we only have to change in one place but now that I think about we would probably have to change all the arguments in places where this method is used anyway 😄

TomMcL (approved):
lgtm

PR #236, opened 2022-11-11
feat: Updating version update script

TomMcL:
### Description

Updating generation script for new version updates to include additional environment variables which were added recently.

### Testing

Ran the script, observed .env file looks correct

### Breaking Changes

None


cdummett (approved):
LGTM

PR #239, opened 2022-11-14
feat: update liquidity offset calculations

cdummett:
### Description
Updates the `ExponentialShapedMarketMaker` to calculate it's liquidity offsets from an estimated mid-price rather than the outdated mid-price from before order amendments (current implementation).

The agent has to calculate an estimate to avoid putting calls to `wait_for_datanode_sync` and data-node queries inside agent `step` methods.

The estimated mid-price is calculated from the average of the MM's best-bid and best-ask and therefore assumes the MM is providing the market best-bid and market-best ask (valid assumption for most market-sim cases). The agent is 

Change fixes the full vega wallet raising errors for negative offsets during large price-movements.

### Testing
Passing all tests locally.

### Breaking Changes
None

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -1274,12 +1274,17 @@ class ExponentialShapedMarketMaker(ShapedMarketMaker):
         buy_specs = [["PEGGED_REFERENCE_BEST_BID", 5, 1]]
         sell_specs = [["PEGGED_REFERENCE_BEST_ASK", 5, 1]]
 
+        if (self.curr_asks is not None) and (self.curr_bids is not None):
+            est_mid_price = (self.curr_bids[0].price + self.curr_asks[0].price) * 0.5
+        elif state is not None:
+            est_mid_price = state.market_state[self.market_id].midprice
+
         if self.curr_asks is not None:
             next_ask_step = self.curr_asks[-1].price + self.tick_spacing
             sell_specs = [
                 [
                     "PEGGED_REFERENCE_MID",
-                    next_ask_step - state.market_state[self.market_id].midprice,
+                    next_ask_step - est_mid_price,
                     1,
                 ]
             ]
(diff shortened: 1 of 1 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -1274,12 +1274,17 @@ def _liq_provis(self, state: VegaState) -> LiquidityProvision:
         buy_specs = [["PEGGED_REFERENCE_BEST_BID", 5, 1]]
         sell_specs = [["PEGGED_REFERENCE_BEST_ASK", 5, 1]]
 
+        if (self.curr_asks is not None) and (self.curr_bids is not None):
+            est_mid_price = (self.curr_bids[0].price + self.curr_asks[0].price) * 0.5
+        elif state is not None:
```
TomMcL, line 1277: This needs to either just be an `else` or have a third option

TomMcL (changes requested):

vega_sim/scenario/common/agents.py
cdummett: The method gets called in cases where all inputs are none. Had as just an else at first but raises an error in the case when `state is None` on initialisation.

I have moved the default `buy_specs` and `sell_specs` to the `else` section. Think this behaviour makes sense?

```
if (self.curr_asks is not None) and (self.curr_bids is not None):
    est_mid_price = (self.curr_bids[0].price + self.curr_asks[0].price) * 0.5
elif state is not None:
    est_mid_price = state.market_state[self.market_id].midprice
else:
    buy_specs = [["PEGGED_REFERENCE_BEST_BID", 5, 1]]
    sell_specs = [["PEGGED_REFERENCE_BEST_ASK", 5, 1]]
```

TomMcL: I see from matching up the if conditions that we probably can't hit the condition, but I don't like leaving a variable uninitialised (as it would be in the `else` here) when it's used later. In this case, we can probably just set `est_mid_price = None` in the else block (later code could handle the case where it is None, but can't handle the case where it doesn't exist at all).

cdummett: Ah fair point, I agree. Have made that change 👍

TomMcL (approved):
lgtm

PR #240, opened 2022-11-15
feat: Switch tx store to always on + update vega

TomMcL:
### Description

Switching transaction store creation to always on, as performance impact seems minimal

### Testing

Ran tests


cdummett (approved):
LGTM

PR #242, opened 2022-11-16
feat: network_service upgrades

cdummett:
## Description
Combo of minor fixes and upgrades to the `network_service.py` module.

#### Features:
- Aligns how `VegaServiceNetwork` manages services with how `VegaServiceNull` manages services.
- Allows `VegaServiceNetwork` to launch a console for the specified network.

#### Fixes: 
- Corrects faulty default `step_bias` value in the `MarketOrderTrader`.
- Adds a missing `initial_mint` arg to the `ConfigurableMarketManager` agent.
- Updates the `Network` enum to match the latest network config files.

#### Refactors:
- Moves `LivePrice` iterator class to somewhere more sensible. 

## Testing
Passing all tests locally.

### DIFF
--- a/vega_sim/network_service.py
+++ b/vega_sim/network_service.py
@@ -25,60 +25,28 @@ Todo:
     • Test all market sim agents are compatible.
 """
 
-import subprocess
-import logging
 import toml
+import time
+import signal
+import logging
+import tempfile
+import subprocess
+import webbrowser
+import multiprocessing
 
-from os import getcwd, path
+from typing import Optional
 
+from os import getcwd, path, environ
+
+from vega_sim import vega_bin_path
 from vega_sim.service import VegaService
 from vega_sim.wallet.base import Wallet
 from vega_sim.wallet.vega_wallet import VegaWallet
+from vega_sim.null_service import find_free_port, _popen_process
 
 from vega_sim.constants import DATA_NODE_GRPC_PORT
 
-
-def start_wallet_service(
-    network: str,
-    automatic_consent: bool = True,
-    no_version_check: bool = False,
-) -> subprocess.Popen:
-    """Starts a subprocess running a wallet service connected to a network.
-    Function uses the local install of vega wallet to start a wallet service
-    connected to the specified network. By default the automatic-consent flag
-    is enabled and the no-version-check flag is disabled.
-
-    Args:
-        network (str):
-            Name of the network config file to use to run the service.
-        automatic_consent (bool, optional):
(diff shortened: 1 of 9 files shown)


vega_sim/network_service.py
```diff
@@ -228,6 +299,13 @@ def data_node_grpc_url(self) -> str:
             self._data_node_grpc_url = f"{url.split(':')[0]}:{DATA_NODE_GRPC_PORT}"
         return self._data_node_grpc_url
 
+    @property
+    def data_node_query_url(self) -> str:
+        if self._data_node_query_url is None:
+            url = self.network_config["API"]["GraphQL"]["Hosts"][1]
```
TomMcL, line 308: I know we're updating this so probably no need to actually change here, but wouldn't this be safer as [0] for the case where there's only one?

vega_sim/scenario/constants.py
```diff
@@ -1,8 +1,9 @@
...
 class Network(Enum):
-    NULLCHAIN = "nullchain"
-    STAGNET1 = "vegawallet-stagnet1"
-    STAGNET3 = "stagnet3"
-    FAIRGROUND = "fairground"
+    nullchain = "nullchain"
+    stagnet1 = "stagnet1"
+    stagnet2 = "stagnet2"
+    stagnet3 = "stagnet3"
+    fairground = "fairground"
```
TomMcL: Why are these lowercase now?

vega_sim/network_service.py
cdummett: Definitely right, it's a legacy bit from when I needed a quick fix for node [0] being down!

It is replaced in the next PR.

vega_sim/scenario/constants.py
cdummett: So the member refer to the directory name and the value refers to the name of the `.toml` config file in that directory. Before it assumed the directory name and config file name both matched the members value.

This is a workaround for the cases where the network file is prefixed with `vegawallet` whilst the directory isn't. Have just realised `stagnet2` is wrong. Will push a fix now.

TomMcL: I would say these should ideally stay as capitals since they are enum constants. If we need to map to something else then an additional mapping dictionary would be best.

cdummett: Sounds good, have reverted back to capitalised Enum members 👍

TomMcL (approved):
LGTM assuming tests pass

PR #243, opened 2022-11-17
feat: network error handling

cdummett:
### Description
PR adds error handling capabilities to the `VegaServiceNetwork` and `NetworkEnvironment` classes.

- Environment able to skip agents if an error is raised by their actions.
- Environment able to check the health of data nodes and switch node if necessary.

Functionality allows users to deploy scenarios to networks which won't be crashed by non critical errors.

### Testing
Passing all tests locally.

### DIFF
--- a/vega_sim/environment/environment.py
+++ b/vega_sim/environment/environment.py
@@ -417,7 +417,7 @@ class NetworkEnvironment(MarketEnvironmentWithState):
 
     def _run(self, vega: VegaServiceNetwork):
         state_values = []
-        
+
         # Initial datanode connection check
         vega.check_datanode(raise_on_error=self.raise_datanode_errors)
 
--- a/vega_sim/environment/environment.py
+++ b/vega_sim/environment/environment.py
@@ -20,6 +20,7 @@ For examples of this setup, see examples/agent_market.
 
 """
 
+import grpc
 import time
 import datetime
 import logging
@@ -384,6 +385,8 @@ class NetworkEnvironment(MarketEnvironmentWithState):
             Callable[[VegaServiceNetwork, List[Agent]], Any]
         ] = None,
         state_extraction_freq: int = 10,
+        raise_datanode_errors: Optional[bool] = True,
+        raise_step_errors: Optional[bool] = True,
     ):
         super().__init__(
             agents=agents,
@@ -398,6 +401,9 @@ class NetworkEnvironment(MarketEnvironmentWithState):
             state_func if state_func is not None else self._default_state_extraction
         )
 
+        self.raise_datanode_errors = raise_datanode_errors
+        self.raise_step_errors = raise_step_errors
+
     def run(self):
         if self._vega is None:
             with VegaServiceNetwork(
@@ -409,8 +415,11 @@ class NetworkEnvironment(MarketEnvironmentWithState):
         else:
             return self._run(self._vega)
 
-    def _run(self, vega):
+    def _run(self, vega: VegaServiceNetwork):
         state_values = []
+
+        # Initial datanode connection check
+        vega.check_datanode(raise_on_error=self.raise_datanode_errors)
 
         # Initialise agents without minting assets
         for agent in self.agents:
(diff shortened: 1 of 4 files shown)


vega_sim/environment/environment.py
```diff
@@ -446,4 +458,11 @@ def step(self, vega: VegaServiceNetwork) -> None:
             if self.random_agent_ordering
             else self.agents
         ):
-            agent.step(state)
+            try:
+                agent.step(state)
+            except:
+                msg = f"Agent '{agent.key_name}' failed to step."
+                if raise_on_error:
+                    raise RuntimeError(msg)
```
TomMcL: We should catch and re-raise the error the `step` threw here or we'll hide it (so just use `except Exception as e` then raise that `e`)

vega_sim/environment/environment.py
```diff
@@ -437,7 +449,7 @@ def _run(self, vega):
             state_values.append(self._state_extraction_fn(vega, self.agents))
             return state_values
 
-    def step(self, vega: VegaServiceNetwork) -> None:
+    def step(self, vega: VegaServiceNetwork, raise_on_error) -> None:
```
TomMcL: Why is this needed vs just checking `self.raise_step_errors`?

vega_sim/network_service.py
```diff
@@ -337,6 +342,83 @@ def core_client(self) -> None:
...
+
+        try:
+            self.ping_datanode()
+            return
+
+        except grpc.FutureTimeoutError:
+            # Unable to connect to datanode
+            msg = f"Connection to endpoint {self._data_node_grpc_url} timed out."
+            if raise_on_error:
+                raise RuntimeError(msg)
```
TomMcL: Same as above, raise the actual error thrown (Although we're at least raising a specific error here)

vega_sim/network_service.py
```diff
@@ -337,6 +342,83 @@ def core_client(self) -> None:
...
+        attempts = 0
+        while attempts != max_attempts:
+            attempts += 1
+
+            try:
+                self.datanode_index += 1
+                if self.datanode_index >= len(
+                    self.network_config["API"]["GRPC"]["Hosts"]
+                ):
+                    self.datanode_index = 0
```
TomMcL: Let's shorten this to `datanode_index = (datanode_index + 1) % len(self.network_config["API"]["GRPC"]["Hosts"])`

TomMcL: Actually, let's use `itertools.cycle(self.network_config["API"]["GRPC"]["Hosts"])` then you can just call `next` on it until the number of tries has been reached

vega_sim/network_service.py
```diff
@@ -337,6 +342,83 @@ def core_client(self) -> None:
...
+                )
+
+                return
+
+            except grpc.FutureTimeoutError:
+                logging.warning(
+                    f"Connection to endpoint {self._data_node_grpc_url} timed out."
+                )
+
+        raise grpc.FutureTimeoutError
```
TomMcL: Exception here should probably just be an `Exception("Data node connection failed blah")` or similar to be a little more obvious about the issue (I'm off and on about whether recommended practice is creating a custom error class. It used to be but I think people like it less now)

vega_sim/network_service.py
```diff
@@ -337,6 +342,83 @@ def core_client(self) -> None:
...
+            # Unable to connect to datanode
+            msg = f"Connection to endpoint {self._data_node_grpc_url} timed out."
+            if raise_on_error:
+                raise RuntimeError(msg)
+            else:
+                logging.warning(msg)
+                self.switch_datanode()
+
+        except grpc._channel._InactiveRpcError:
+            msg = f"Connection to endpoint {self._data_node_grpc_url} in active."
```
TomMcL: inactive

TomMcL (changes requested):

TomMcL (approved):
lgtm

PR #244, opened 2022-11-18
chore: Remove step lengths from scenario tests

TomMcL:
### Description

Removing some kwargs which are only really useful when displaying results in Console and slow runs down

### Testing

Ran tests

### Breaking Changes

None


cdummett (approved):
LGTM

PR #246, opened 2022-11-21
feat: default price process fn

cdummett:
### Description
Adds functionality to the `test_parameter_runner.py` script which overrides the configured `price_process_fn` with a `random_walk` model which is used for integration tests.

To achieve the above, the following updates have been made:

1. All scenarios have a `num_steps` arg and attribute
2. All scenarios have a `price_process_fn` arg and attribute
3. All scenarios have a `_generate_price_process` method which can be called if `price_process_fn` is `None`

PR also deprecates the `RW_model` and `GBM_models` used within `vega_sim` (replaced with `random_walk`) and deprecates the `Network` and `Fairground` scenarios (as network scenarios moved to the `devops` dir).

### Testing
Passing all tests locally.


tests/integration/test_parameter_runner.py
```diff
@@ -15,6 +16,13 @@ def test_parameter_runner(experiment_to_run: experiment.SingleParameterExperimen
         experiment_to_run.runs_per_scenario = (
             1  # Likely only need to test the one run per scenario
         )
+
+        experiment_to_run.scenario.price_process_fn = lambda: random_walk(
+            num_steps=experiment_to_run.scenario.num_steps,
+            starting_price=1000,
+            sigma=0.1,
+        )
```
TomMcL, line 24: Hmm, I'm not sure whether we want to blindly override every scenario with the same random walk config ultimately, but it's perhaps ok for now. We should probably at some point separate out the concept of testing the parameter runners work and testing specific scenarios in any case

TomMcL (approved):
lgtm

tests/integration/test_parameter_runner.py
cdummett: I agree, I think we probably need one test per scenario, and then maybe two tests for the parameter runner (network parameter test and market parameter test).

PR #251, opened 2022-11-22
fix: correct random_walk logic

cdummett:
### Description
Running a parameter test:

`python -m vega_sim.parameter_test.run -c TauScaling_Curve
`

Would throw this exception:

`Exception: Negative price generated with current random seed. Please try another or don't specify one
`

Error thrown as the loop calculating the next random price was not long enough, i.e. `S[-1]` left as `0` so the exception raised.

PR fixes `random_walk`.

### Testing
Passing all tests locally.


Jiajia-Cui (approved):
LGTM

PR #252, opened 2022-11-22
feat: tidy up parameter changing agents

TomMcL:
### Description

Remove parameter changing from a few places to avoid hidden updates and make tests more uniform

### Testing

Ran several scenarios, trading still looked reasonable and didn't drop into permanent auction.

### Breaking Changes

None


Jiajia-Cui (approved):
LGTM

PR #255, opened 2022-11-22
feat: CurvedMarketMaker order scaling

### LINKED ISSUE #253: Move CurveMarketMaker to calculate what its liquidity provision requirement is and scale to that amount
Currently the CurveMarketMaker will provide a set amount of liquidity whatever its commitment says it should be (which is topped up by the pegged orders if this happens to not be enough. This wouldn't make much sense for a real market maker as if they wanted to provide additional liquidity they could raise their stake too and receive a higher proportion of the fees.

As such, we should look into directly calculating what the stake requirement is and scaling the shape to match that as closely as possible, with the pegged orders making up what should be a fairly small remaining difference at the end.

These calculations can probably be taken fairly directly from a similar set [here](https://github.com/MM0819/vega-market-maker/blob/main/src/main/java/com/vega/protocol/utils/QuantUtils.java#L24)

cdummett:
### Description
PR adds ability for `CurvedMarketMaker` to scale its orders so its `instantaneous_liquidity` (from persistent orders) roughly equals its `liquidity_obligation`. This change allows market-maker to maximise fees received from it's liquidity provision.

### Testing
Passing all tests locally.

### Breaking Changes

Following unused parameters deprecated from `CurvedMarketMaker`:

- `base_order_size`
- `max_order_size`

Configured scenarios containing a `ExponentialShapedMarketMaker` using these arguments will break. To fix, remove arguments from agent or scenario configuration.


vega_sim/scenario/common/agents.py
```diff
@@ -1125,6 +1129,90 @@ def step(self, vega_state: VegaState):
                 key_name=self.key_name,
             )
 
+    def _scale_orders(self, buy_shape: List[MMOrder], sell_shape: List[MMOrder]):
+
+        est_midprice = 0.5 * (buy_shape[0].price + sell_shape[0].price)
```
davidsiska-vega: I know it's only used to estimate liquidity but you shouldn't need mid. You can always use just the offset. The probability of trading is (by our definition) 1/2 between best bid and ask and then decays using the prob of trading function.

vega_sim/scenario/common/agents.py
```diff
@@ -1125,6 +1129,90 @@ def step(self, vega_state: VegaState):
...
+                self._calculate_liquidity(
+                    side=vega_protos.SIDE_SELL,
+                    orders=sell_shape,
+                    midprice=est_midprice,
+                ),
+            ]
+        )
+
+        # Compute the scaling factor
+        stake_to_ccy_siskas = self.vega.get_network_parameter(
```
davidsiska-vega: Would getting this at every step of the agent slow things down? Maybe we can just get it once upfront? Though this is of course safer if it somehow gets changed half-way through a run.

davidsiska-vega (commented):
Looks good to me, left a couple of small comments.

vega_sim/scenario/common/agents.py
cdummett: I think it would slow things down a little, mostly have it in as like you say in case it get's updated half-way through a run (most likely if `devops` are using this on a network).

cdummett: I'll add a flag so we can change the behaviour:

- When running a sim (parameter unlikely to change) - don't update each step
- When deploying to a network (parameter may change) - check every step (or every n steps)

vega_sim/scenario/common/agents.py
cdummett: Ah nice thanks, that makes sense, I'll make that change

vega_sim/scenario/common/agents.py
TomMcL: If you call it something along the lines of "sync_state_every_n_steps" (but maybe more concise 😄 ) you can make it a general function which it calls and we can add other things to in future too (might be able to put it in the base class too)

vega_sim/api/data.py
```diff
@@ -427,6 +427,36 @@ def best_prices(
...
+    ]
+    upper_bounds = [
+        price_monitoring_bound.max_valid_price
+        for price_monitoring_bound in mkt_data.price_monitoring_bounds
+    ]
+
+    return (
+        num_from_padded_int(max(lower_bounds), mkt_price_dp),
+        num_from_padded_int(min(upper_bounds), mkt_price_dp),
+    )
```
TomMcL, line 457: Add a test for this please

vega_sim/quant/quant.py
```diff
@@ -0,0 +1,34 @@
...
+def probability_of_trading(
+    tau,
+    mu,
+    sigma,
+    lower_bound: float,
+    upper_bound: float,
+    best_price,
+    price,
+    side,
+):
```
TomMcL: Docstring and type hints

vega_sim/quant/quant.py
```diff
@@ -0,0 +1,34 @@
...
+
+    stdev = sigma * sqrt(tau)
+    m = log(best_price) + (mu - 0.5 * sigma * sigma) * tau
```
TomMcL: `sigma ** 2`

vega_sim/quant/quant.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if price < lower_bound or price > upper_bound:
+        return 0
+
+    min = cdf(m, stdev, lower_bound)
+    max = cdf(m, stdev, upper_bound)
```
TomMcL: Should probably use scipy stats here rather than reimplementing locally

vega_sim/scenario/common/agents.py
cdummett: Sounds good, does this make sense to do this outside the agent in the environment so we can pass to every agent?

Only issue would be some data is required every step (market data etc.) and some data is only required occasionally (market info, network parameters etc.)

vega_sim/scenario/common/agents.py
```diff
@@ -1125,6 +1129,90 @@ def step(self, vega_state: VegaState):
...
+        sigma = market_info.tradable_instrument.log_normal_risk_model.params.sigma
+
+        tau_scaling = self.vega.get_network_parameter(
+            "market.liquidity.probabilityOfTrading.tau.scaling",
+            "float",
+        )
+        min_probability_of_trading = self.vega.get_network_parameter(
+            "market.liquidity.minimum.probabilityOfTrading.lpOrders",
+            "float",
+        )
```
TomMcL: These can go in that state fn too

vega_sim/scenario/common/agents.py
```diff
@@ -1125,6 +1129,90 @@ def step(self, vega_state: VegaState):
...
+                        lower_bound=min_valid_price,
+                        upper_bound=max_valid_price,
+                        best_price=midprice,
+                        price=price,
+                        side=side,
+                    ),
+                ]
+            )
+
+            provided_liquidity += vol * price * p
```
TomMcL, line 1210: Instinctively, this loop (and so ultimately the `probability_of_trading` function) can almost certainly be vectorised and run through numpy functions. Our lists probably aren't long enough that it's worth doing now but if you ever want to have a shot then give it a go 😄

TomMcL (changes requested):

vega_sim/scenario/common/agents.py
TomMcL: No I think there's some stuff that's useful to specific agents only so can be inside the step. Though we do also want to keep general market data in the VegaState object where possible to pass to every agent, but I think network parameters are specific enough to load locally

vega_sim/quant/quant.py
cdummett: Have switched over but am getting eeeever so slightly different probabilities from Lewis units tests.

- Scipy:  0.9997857769969356
- Local implementation: 0.9997857769969329

Am going to assume this is okay and update our unit tests to reflect the scipy values...?

TomMcL: Yeah when the 15th significant figure is on our list of biggest accuracy issues I'll be happy :).

Side note: Are his/our unit tests comparing exact values of floats? They shouldn't be doing that in any case (for this exact reason)

cdummett: Floats, I have updated our tests a little so we just check all the limits.

vega_sim/scenario/common/agents.py
cdummett: Thanks yeah I will definitely come back to this at some point!

vega_sim/environment/agent.py
```diff
@@ -47,14 +50,32 @@ def __init__(
...
-    def step(self, vega: VegaService):
-        pass
+        self.step_count = 0
+        self.market_params_update_freq = market_params_update_freq
+        self.network_params_update_freq = network_params_update_freq
+
+        self.market_params = None
+        self.network_params = (
+            dict.fromkeys(network_params, None) if network_params is not None else None
+        )
```
TomMcL: This feels perhaps a little too specific to our immediate use case to me. I'd probably stick to it just being a generic "state_update_freq" and then have individual agents update network params within that if they feel like it. But maybe params updates are general enough they deserve to have their own thing I'm not sure.

In any case I would certainly aggregate into one function at least, don't think we need to update them at different frequencies.

vega_sim/environment/agent.py
```diff
@@ -67,12 +88,31 @@ def initialise(self, vega: VegaService, create_wallet: bool = True):
         else:
             self.vega.login(name=self.wallet_name, passphrase=self.wallet_pass)
 
+    def _update_market_params(self):
+
+        if self.market_params_update_freq is None:
+            return
+
+        if self.step_count % self.market_params_update_freq == 0:
```
TomMcL: You can drop the above check and just do `if self.market_params_update_freq and self.step_count % self.market_params_update_freq == 0`

vega_sim/environment/agent.py
cdummett: I think you might be right, I was going for a super general function which could be used by any agent, but it probably overcomplicates things. 

Will move things around to match your suggestion  😄

TomMcL (approved):
lgtm

PR #260, opened 2022-11-24
feat: add one_off_transfers method

### LINKED ISSUE #264: Add `transfer` method to `VegaService`

cdummett:
### Description
Adds a `one_off_transfer` method to `VegaService`.

### Testing
Integration tests added to check following cases:

- If no delay: funds taken from account and transferred to account instantly.
- If delay: funds taken from account instantly but only transferred to account on deliver time.

Passing all tests locally.

### DIFF
--- a/tests/integration/utils/fixtures.py
+++ b/tests/integration/utils/fixtures.py
@@ -128,12 +128,6 @@ def build_basic_market(
     )
     vega.wait_for_total_catchup()
 
-    vega.update_network_parameter(
-        proposal_wallet=MM_WALLET.name,
-        parameter="transfer.fee.factor",
-        new_value="0",
-    )
-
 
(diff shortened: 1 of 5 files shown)


tests/integration/utils/fixtures.py
```diff
@@ -128,6 +128,12 @@ def build_basic_market(
     )
     vega.wait_for_total_catchup()
 
+    vega.update_network_parameter(
+        proposal_wallet=MM_WALLET.name,
+        parameter="transfer.fee.factor",
+        new_value="0",
```
TomMcL: If this is non-zero normally we can probably just change it in `genesis.json`, seems like we probably want it as 0 by default in market sim

TomMcL (approved):
lgtm

PR #270, opened 2022-11-29
feat: Adding arg to learning script + fixing bounds issue

TomMcL:
### Description

Adding argument to learning script and making bounds function handle when there are none

### Testing

Tested script locally

### Breaking Changes

Anything calling the learning script (nothing yet) would need to now pass a number of learning iterations


cdummett (approved):
LGTM! Tested locally no issues. Still issues running in Docker for me but the configurable runs works:

```
vega-market-sim caed$ sh scripts/run-docker-learning.sh 5
```

PR #272, opened 2022-12-01
feat: transfer monitoring

cdummett:
### Description
PR adds a separate monitoring stream for transfers.

Monitoring stream will be used to query transfers `to` a `party_id` (required for the hedged market maker to be able to track inbound transfers which have been accepted but the deliver time has not been reached).

### Testing

Unit tests added for new `data` and `data_raw` methods and integration tests updated for new `service` methods.

Passing all tests locally 👍 

### Breaking Changes
None

### DIFF
--- a/vega_sim/api/data.py
+++ b/vega_sim/api/data.py
@@ -7,6 +7,7 @@ import vega_sim.api.data_raw as data_raw
 import vega_sim.grpc.client as vac
 import vega_sim.proto.data_node.api.v2 as data_node_protos_v2
 import vega_sim.proto.vega as vega_protos
+import vega_sim.proto.vega.events.v1.events_pb2 as events_protos
 from vega_sim.api.helpers import num_from_padded_int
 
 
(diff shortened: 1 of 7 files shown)


vega_sim/api/data.py
```diff
@@ -7,6 +7,7 @@
 import vega_sim.grpc.client as vac
 import vega_sim.proto.data_node.api.v2 as data_node_protos_v2
 import vega_sim.proto.vega as vega_protos
+import vega_sim.proto.vega.events.v1.events_pb2 as events_protos
```
davidsiska-vega, line 10: I am just wondering whether this is the v1 apis that are getting deprecated, or do I have things mixed up?

davidsiska-vega (approved):
Once question left, but LGTM.

vega_sim/api/data.py
TomMcL: I believe should fine, it's just the data_node part which is changing (so far as I'm aware)

cdummett: I thought it was just v1 datanode APIs that are getting deprecated, as far as I know there is no v2 for the other gRPC methods..?

vega_sim/api/data.py
```diff
@@ -770,6 +813,44 @@ def _order_gen(
...
+                        )
+                    yield _transfer_from_proto(
+                        transfer=transfer,
+                        asset_decimals=asset_dp[transfer.asset],
+                    )
+        except Exception as _:
+            logger.info("Transfer subscription closed")
+            return
+
+    return _transfer_gen(transfer_stream=transfer_stream)
```
TomMcL, line 851: Possible at some point we'll need a single unified listener rather than one per event, but this seems reasonable for now

TomMcL (approved):

PR #278, opened 2022-12-05
fix: limit max order size

cdummett:
### Description
RL scenario was causing the `ShapedMarketMaker` agent to post large orders which were rejected by core.

PR adds an arg which limits each order to a maximum size.

### Testing
Passing all tests locally.

### Breaking Changes
None.

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -966,6 +966,7 @@ class ShapedMarketMaker(StateAgentWithWallet):
         orders_from_stream: Optional[bool] = True,
         state_update_freq: Optional[int] = None,
         safety_factor: Optional[float] = 1.2,
+        max_order_size: Optional[float] = 10000,
     ):
         super().__init__(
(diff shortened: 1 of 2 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -966,6 +966,7 @@ def __init__(
         orders_from_stream: Optional[bool] = True,
         state_update_freq: Optional[int] = None,
         safety_factor: Optional[float] = 1.2,
+        max_order_size: Optional[float] = 10000,
```
TomMcL: `Optional` means "can be None" rather than "is a kwarg" (Why they called it that when kwargs are already referred to as optional arguments we shall never know..). I don't think this can be `None` right?

cdummett: Ah thanks, I didn't know that. Yeah that can't be `None` have pushed a change.

TomMcL (approved):

PR #282, opened 2022-12-14
feat: add hedged market scenario

### LINKED ISSUE #258: feat: HedgedMarketMaker scenario
### Goal:
Create a scenario to mimic a Market-Maker (MM) who wants to follow a hedging strategy; i.e. making a market on a Vega network (profiting of fees) and hedging its position on a highly liquid external exchange.

Viability of strategy relies on the MM being able to move funds between the exchanges fast enough to prevent close-outs from insufficient margins.

### Aim:

Test whether Vega's withdrawal delay is too long or its margin requirements too strict for this strategy to be viable 

#### Markets

Scenario will have two markets:

1.  `MarketA` - represents an expected early Vega market (`price_process` can follow `MarketB`)
2. `MarketB` - represents a highly liquid external market (`price_process` can follow an external source)

The MM will operate in both markets and have a separate key for each market (`KeyA` and `KeyB`). The keys represent the separate accounts on the Vega network and external exchange. The MM will be able to transfer funds between keys but subject to delays. Moving funds from `KeyB` too `KeyA` will have small delays (e.g. `15m`) but moving funds from `KeyA` too `KeyB` will have large delays (e.g. `24h`). This is to mimic Vega's withdrawal delays.

The MM will follow a hedging strategy, posting limit orders on `MarketA`, and hedging its position on `MarketB` (the "external" market). The MMs order depths will be determined such that is it able to profit off the liquidity + maker fees. The MM will then move funds between keys to manage it's margins and prevent close-outs.

#### Agents

In `MarketA` an `InformedTrader` who is aware of the next price will generate a position for the MM. We can run two cases one where the `InformedTrader` is correct (and the MM will have to transfer funds from `KeyB` to `KeyA`, encountering no delays) and one where the `InformedTrader` is incorrect (and the MM will have to transfer funds from `KeyA` to `KeyB`, encountering delays)

### ToDo:

- [x] #264
- [x] #265
- [x] #266
- [x] #267
- [x] #268

### LINKED ISSUE #265: Create the `HedgedMarketMaker` agent

### LINKED ISSUE #266: Add "wrong" functionality to `InformedTrader`

### LINKED ISSUE #267: Create the `HedgedMarketScenario`

cdummett:
#### HedgedMarketMaker agent:

PR adds a `HedgedMarketMaker` (HMM) agent which makes a market in an internal market and hedges its position on an external (liquid) market. The agent uses two separate keys to represent separate accounts and tries to maintain equal leverage between the accounts through transfers.

- The HMM makes a market by setting bid / ask depths to meet a configurable profit margin and submits shaped limit orders beyond these depths (when calculating depths, any fees paid / earned are accounted for).
- The HMM hedges its position by placing a market order on the external market with a volume equal to the difference between the position on the internal and external markets.
- The HMM maintains similar leverage on either market by transferring the difference between accounts on the internal and external markets whenever the difference is above some threshold (the difference accounts for any funds currently locked in transfers).

#### HedgedMarket scenario:

PR adds a `HedgedMarket` scenario which uses two `MarketManager` agents to create an "internal" market (representing a Vega market) and an "external" market (representing an external liquid market).

A `HedgedMarketMaker` makes a market on the internal market and an `ExponentialShapedMarketMaker` makes a market on the external market. `MarketOrderTrader` agents provide low volume trades each step in both markets and a configurable `InformedTrader` submits orders in the internal market. The accuracy of the InformedTrader can be configured to cause the `HedgedMarketMaker` to make loses either on its primary market where it is making a market, or its secondary market where it is hedging its position.

```
$ python -m vega_sim.scenario.adhoc  --console --debug --pause --pause_every_n_steps 240  -s hedged_market
```

#### InformedTrader updates:

`InformedTrader` updated with functionality to "look-ahead" rather than looking at the next step as well as a configurable accuracy which determines whether the `InformedTrader` speculates the correct side to buy / sell

### Testing
Passing all integration tests locally. Passing RL scripts.

### Breaking Changes
None (AFAIK)

### Closes
Closes #258 
Closes #265
Closes #266 
Closes #267


vega_sim/scenario/common/agents.py
```diff
@@ -1500,6 +1502,298 @@ def _calculate_price_volume_levels(
...
+            ]
+        )
+
+        # Calculate the difference between the total balance on the markets
+        delta = (internal_account_balance + internal_transfers_balance) - (
+            external_account_balance + external_transfers_balance
+        )
+
+        # Create a transfer to balance the internal and external accounts
+        if abs(delta) < 500:
```
TomMcL: This could probably be configurable to set how often they rebalance

vega_sim/scenario/common/agents.py
```diff
@@ -1828,21 +2168,63 @@ def step(self, vega_state: VegaState):
...
+                fill_or_kill=False,
+                key_name=self.key_name,
+            )
+            return ITOrder(side=side, size=size)
+
+        except OrderRejectedError:
+            logger.debug("Order rejected")
+            return None
+
+    def _settle_order(self, order: ITOrder):
```
TomMcL: Interesting setup. So 'settle' here means the agent closing out the amount of position it took from the matching original position? Seems like a good plan. A couple of changes I'd probably make would be making the queue length configurable (or maybe even make a max 'allowed abs position' before running the closeouts too?) and rename to something slightly clearer about what's going on, I'm thinking `_close_positions` or something?

cdummett: > So 'settle' here means the agent closing out the amount of position it took from the matching original position?

Yeah exactly, very happy to change the name of the method, `_close_positions` sounds good.

> couple of changes I'd probably make would be making the queue length configurable

So the queue length is configured through the `lookahead` arg, as we only get an order from the queue when the queue is full we know when we do the correct number of steps has passed (why its also important to put a "fake" order into the queue even if no order is placed).

This does break however if a market goes into auction because obvs market orders no longer work... open to ideas on this, was maybe thinking once it exits auctions it closes its position and empties the queue.

> (or maybe even make a max 'allowed abs position' before running the closeouts too?)

I like this will add this in 👍

TomMcL: Makes sense on `lookahead`. And yeah seems reasonable to just wait if the market is in auction and close out when it reopens since this agent isn't really trying to do anything special whilst in auction.

cdummett:
@TomMcL Jenkins failing to build with the correct tag ([Git Build Data](https://jenkins.ops.vega.xyz/job/common/job/vega-market-sim/1074/git-3/)) but all tests passing locally.

vega_sim/scenario/hedged_market_maker/agents.py
```diff
@@ -0,0 +1,17 @@
...
+PartyConfig = namedtuple("PartyConfig", ["wallet_name", "wallet_pass", "key_name"])
+
+MARKET_CREATOR = PartyConfig("aux_parties", "pass", "Market Creator")
+MARKET_SETTLER = PartyConfig("aux_parties", "pass", "Market Settler")
+
+AUCTION_PASS_BID = PartyConfig("aux_parties", "pass", "Auction Pass (Bid)")
+AUCTION_PASS_ASK = PartyConfig("aux_parties", "pass", "Auction Pass (Ask)")
+
+INT_MARKET_MAKER_KEY_A = PartyConfig("market_maker", "pass", "Key A (Internal)")
+INT_MARKET_MAKER_KEY_B = PartyConfig("market_maker", "pass", "Key B (External)")
```
TomMcL: Is there a reason to put these in different wallets?

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+            vega.create_wallet(
+                name=self.wallet_name,
+                passphrase=self.wallet_pass,
+                key_name=self.external_key_name,
+            )
+        if mint_wallet:
+            vega.mint(
+                wallet_name=self.wallet_name,
+                asset=self.asset_id,
+                amount=self.initial_asset_mint,
```
TomMcL: We might want to control how much is minted in each side at some point, but this seems fine for now

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+            self.int_liq_fee = (
+                float(int_market_info.fees.factors.liquidity_fee)
+                if int_market_info is not None
+                else 0
+            )
+            self.int_inf_fee = (
+                float(int_market_info.fees.factors.infrastructure_fee)
+                if int_market_info is not None
+                else 0
+            )
```
TomMcL: Could we collapse all of these into a single `if int_market_info is not None:` block? And is there ever a time where `self.market_id` is not None and `int_market_info` still is? If not, lets roll all four into one

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+
+    def _optimal_strategy(self, current_position, current_step):
+
+        ext_best_bid, ext_best_ask = self.vega.best_prices(
+            market_id=self.external_market_id
+        )
+
+        int_fee = self.int_mkr_fee + self.int_liq_fee
```
TomMcL, line 1616: I think this is only going to be true in the case where the MM is the *only* MM so that all liquidity fees go to them. Worth being aware of if so

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+        ext_fee = self.ext_mkr_fee + self.ext_liq_fee + self.ext_inf_fee
```
TomMcL, line 1617: Technically I guess we want this to be flexible, but should be able to do that through tweaking the liquidity fee on the market

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+
+        # Determine the delta between the position on the internal and external market
+
+        internal_position = self.vega.positions_by_market(
+            wallet_name=self.wallet_name,
+            market_id=self.market_id,
+            key_name=self.key_name,
+        )
+        current_int_position = (
+            float(internal_position[0].open_volume) if internal_position != [] else 0
```
TomMcL: Is this not already a float?

vega_sim/scenario/common/agents.py
```diff
@@ -1473,6 +1475,301 @@ def _calculate_price_volume_levels(
...
+        )
+        external_position = self.vega.positions_by_market(
```
TomMcL, line 1643: I would be nice to generify the `positions_by_market` function at some point to be able to pass `None` as the market_id and get everything back. Some of our later data access functions handle this. Doesn't need doing now though, two calls isn't too bad :)

vega_sim/scenario/hedged_market_maker/scenario.py
```diff
@@ -0,0 +1,333 @@
...
+        int_it_proportion: float = 0.1,
+        int_it_accuracy: float = 1.0,
+        int_it_lookahead: int = 30,
+        random_agent_ordering: bool = False,
+        transactions_per_block: int = 1000,
+        state_extraction_fn: Optional[
+            Callable[[VegaServiceNull, List[Agent]], Any]
+        ] = None,
+        state_extraction_freq: int = 1,
+        pause_every_n_steps: Optional[int] = None,
```
TomMcL, line 189: If you're feeling brave, we should probably start trying to be better at actually documenting what the million parameters passed into these scenarios actually do 😄

vega_sim/scenario/registry.py
```diff
@@ -156,4 +157,18 @@
...
+            product_id="ETH-USD",
+            granularity=Granularity.MINUTE,
+            start="2022-06-12 17:00:00",
+            end="2022-06-13 17:00:00",
+        ).values,
+        int_lock=3 * 60 * 60,
+        ext_lock=5 * 60,
+    ),
 }
+688
```
TomMcL: 688?

vega_sim/scenario/hedged_market_maker/scenario.py
```diff
@@ -0,0 +1,333 @@
...
+        int_random_trader = MarketOrderTrader(
+            wallet_name=INT_RANDOM_TRADER.wallet_name,
+            wallet_pass=INT_RANDOM_TRADER.wallet_pass,
+            key_name=INT_RANDOM_TRADER.key_name,
+            market_name=self.int_name,
+            asset_name=self.asset_name,
+            initial_asset_mint=1e10,
+            buy_intensity=1 * 10 ** (self.int_pdp),
+            sell_intensity=1 * 10 ** (self.int_pdp),
+            base_order_size=1 * 10 ** -(self.int_pdp),
```
TomMcL, line 398: Might need to make these configurable again at some point

vega_sim/scenario/hedged_market_maker/agents.py
cdummett: Only to make it a bit clearer in the console when switching between market-maker keys, keeps all the auxiliary keys which aren't really related to the scenario hidden.

vega_sim/scenario/common/agents.py
cdummett: Ah that is a very good point, that could be tricky, not sure what gRPC request we could use to estimate ELS either..?

vega_sim/scenario/registry.py
cdummett:
```suggestion
```

cdummett: Thats a typo thanks for spotting

vega_sim/scenario/common/agents.py
cdummett:
```suggestion
           internal_position[0].open_volume if internal_position != [] else 0
```

vega_sim/scenario/hedged_market_maker/agents.py
cdummett: Do you reckon this is a complexity we don't need now we have better logging (i.e. need console less)?

vega_sim/scenario/hedged_market_maker/scenario.py
cdummett: Sounds good, do you think it is worth starting to flesh out each scenario with a README? Detailed enough docstrings could start to get very long for some of these scenarios I think.

vega_sim/scenario/common/agents.py
cdummett: Sounds good, raised an [issue](https://github.com/vegaprotocol/vega-market-sim/issues/285) so I don't forget 😄

vega_sim/scenario/hedged_market_maker/scenario.py
cdummett: I added these agents to provide a small trade at each step to guarantee candle data. Think I left these out to try avoid having too many arguments in the scenario. Happy to add arguments for these in the future though if we see another use for the `MarketOrderTraders`.

vega_sim/scenario/common/agents.py
cdummett: Linked to my comment on configurable args for the MarketOrderTrader but I think this one makes a lot of sense actually. Might be interesting to look at simulations where the accounts are unbalanced.

I'll make the change thanks 👍

vega_sim/scenario/common/agents.py
cdummett: Ah nice yeah you are right thanks, have made that change 👍

vega_sim/scenario/hedged_market_maker/agents.py
TomMcL: More concerned about whether wallet v2 will ever let us create wallets dynamically. Although given we already have custom named wallets here it may be no worse anyway :D

vega_sim/scenario/common/agents.py
TomMcL: After an epoch you can pull the calculated one, but obviously most of our runs don't really go that long! So we could either shorten epochs or just leave it for now 😄. For the purpose of the sim we're actually using this for it's correct anyway. Maybe just spin up an issue for it so we remember. (One half way house might be to just use the ratio of liquidity provision over all liquidity provisions, assuming they'll all continue to exist)

vega_sim/scenario/hedged_market_maker/scenario.py
TomMcL: Possibly a good combination would be a module level docstring giving a broad overview of what the scenario runs and does. That way the docstring for the class itself can be a little more terse in how it refers to what args do.

vega_sim/scenario/hedged_market_maker/scenario.py
TomMcL: That makes sense, forgot about that purpose for them! Probably no need to have them at all configurable no

vega_sim/scenario/hedged_market_maker/agents.py
cdummett: Ah that makes sense, I'll condense it all into one wallet and just follow that for any future scenarios 😄

vega_sim/scenario/common/agents.py
cdummett: Sounds good thanks, have raised this issue [here](https://github.com/vegaprotocol/vega-market-sim/issues/286) 👍

TomMcL (approved):

PR #283, opened 2022-12-19
feat: remove prob trading changes

TomMcL:
### Description

Removing probability of trading from the curve market maker's liquidity provision calculations now that it is no longer used in core.

### Testing

Ran historic simulation test, confirmed that liquidity provisions of sensible size were still being posted.

### Breaking Changes

None


Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
         timeout(time: 50, unit: 'MINUTES')
     }
     parameters {
-        string( name: 'VEGA_VERSION', defaultValue: '443af04ef30fd60086d870eff961916f507c68d6',
+        string( name: 'VEGA_VERSION', defaultValue: 'e8c50d94062ac44a55976055cf3290861c70672b',
```
cdummett: Ah this explains it 😆

cdummett (dismissed):
LGTM

Jenkinsfile
TomMcL: :+1:

cdummett (approved):

PR #284, opened 2022-12-19
feat: better datanode switching

### LINKED ISSUE #279: Better Node Switching
Currently the `NetworkEnvironment` class pings the currently connected datanode each time step and if a successful response is not received the  `VegaServiceNetwork` attempts to switch to a new datanode.

This does not work for cases where a datanode is still active but starts to fall behind (As in December Incentive, N06 fell behind, bots started to amend 'old' orders, market fell into auction). 

Node switching trigger needs to be improved. i.e. instead of pinging, request datanode time and compare to current system time. If not within acceptable bounds try a new datanode. Ideally want to avoid methods like comparing block-heights which require connections to multiple datanodes.

cdummett:
### Description
Adds simple comparison between datanode time and system time during datanode health check.

Will fix issues with bots connecting to a lagging datanode during incentives and getting stale order data.

### Testing
Passing all tests locally.

### Breaking Changes
None.

### Closes
Closes #279


vega_sim/network_service.py
```diff
@@ -341,17 +341,21 @@ def core_client(self) -> None:
...
 
         If the current data-node connection has timed out or the end-point is
         unresponsive then the service will attempt to establish a connection
         with a new healthy data-node.
 
         Args:
-            max_attempts (int, optional):
-                Maximum number of connection attempts to attempt before raising an
-                error. Defaults to -1 (infinite attempts).
+            max_time_difference (int, optional):
```
TomMcL: Am I missing where this is actually used?

cdummett: Ah it should be passed to `ping_datanode` thanks. Will update now.

TomMcL (approved):

PR #287, opened 2022-12-21
feat: Adding RL test and fixing up scenario

TomMcL:
### Description

Fixing RL scenario to fix in to the new logging setup and adding a small integration test to ensure breaks are noticed

### Testing

Added integration test which was broken and is now fixed

### Breaking Changes

None


cdummett (dismissed):
LGTM - docker build and tests passing 👍

cdummett (approved):

PR #289, opened 2022-12-23
feat: configurable streaming node

### LINKED ISSUE #273: Configurable node for ObserveEventBus
### Description

When running a market-sim scenario on an existing network, the `VegaServiceNetwork` does not have the functionality to listen to a core node. Instead, the bots use the `list_orders` method to get a list of live orders at each time-step. This is unsuitable for markets with a large number of live orders, e.g. [ETHUSD](https://console.fairground.wtf/#/markets/e9ec6d5c46a7e7bcabf9ba7a893fa5a5eeeec08b731f06f7a6eb7bf0e605b829).

Therefore, the functionality to setup an event bus stream listening to a `datanode` needs to be added.

cdummett:
### Description

PR adds functionality to switch whether service listens to a stream event from a core node or a datanode.

PR also updates how `base_orders` is constructed.
- Previously all orders for all markets requested (fewer requests, faster for sims, slower for networks)
- Now only orders for specified markets and parties requested (multiple requests,  slower for sims, faster for networks)

### Testing
Passing all tests locally.

### Breaking Changes
None

### Closes
Closes #273

### DIFF
--- a/devops/run.py
+++ b/devops/run.py
@@ -69,6 +69,7 @@ def main():
             warn_on_raw_data_access=False,
             run_with_console=args.console,
             launch_graphql=args.graphql,
+            start_order_feed=False,
         ) as vega:
 
             scenario.run_iteration(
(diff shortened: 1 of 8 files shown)


TomMcL (approved):
lgtm

PR #290, opened 2023-01-03
feat: add multiple lps visualisation example

cdummett:
### Description
Adds an example which allows the console liquidity provision tab to be validated.

Example creates a market with multiple liquidity providers submitting provisions at different price levels:

- `LP_A`: `commitment=15000` and `fee 0.01`
- `LP_B`: `commitment=15000` and `fee 0.02`
- `LP_C`: `commitment=15000` and `fee 0.03`

The market then exits the opening auction with a trade where `price=1000` and `size=1`.

Traders then place orders to vary the open interest and target stake in the market:

1. `max_oi=200` and `target_stake>10000`:

- Liquidity fee should be `0.01`.
- Provision from `LP_A`, `LP_B`, and `LP_C` should be `STATUS_DEPLOYED`.
- Share split should be `33%`, `33%`, `33%` between `LP_A`, `LP_B`, and `LP_C` respectively.

2. `max_oi=400` and `target_stake>20000`:

- Liquidity fee should be `0.02`.
- Provision from `LP_A`, `LP_B`, and `LP_C` should be `STATUS_DEPLOYED`.
- Share split should be `33%`, `33%`, `33%` between `LP_A`, `LP_B`, and `LP_C` respectively.

3. `max_oi=600` and `target_stake>30000`:

- Liquidity fee should be `0.03`.
- Provision from `LP_A`, `LP_B`, and `LP_C` should be `STATUS_DEPLOYED`.
- Share split should be `33%`, `33%`, `33%` between `LP_A`, `LP_B`, and `LP_C` respectively.

### Testing
Integration test added. Passing all tests.

### Example
```
$ python -m examples.visualisations.multiple_lps --console --debug --pause
```

### DIFF
--- a/examples/visualisations/closed_out.py
+++ b/examples/visualisations/closed_out.py
@@ -131,6 +131,7 @@ if __name__ == "__main__":
                 best_bid_id=best_bid_id,
                 price=price,
                 spread=10,
+                volume=1,
             )
             vega.wait_for_total_catchup()
             vega.wait_fn(60)
--- a/examples/visualisations/loss_socialisation.py
+++ b/examples/visualisations/loss_socialisation.py
@@ -131,6 +131,7 @@ if __name__ == "__main__":
                 best_bid_id=best_bid_id,
                 price=price,
                 spread=10,
+                volume=1,
             )
             vega.wait_for_total_catchup()
             vega.wait_fn(60)
--- a/examples/visualisations/multiple_lps.py
+++ b/examples/visualisations/multiple_lps.py
@@ -144,7 +144,7 @@ if __name__ == "__main__":
         vega.wait_for_total_catchup()
         input(
             "Pause simulation to allow UI to be inspected. "
-            + "Provision from LP_A should be deployed."
+            + "Fee should be updated to 0.01 (1.00%)"
         )
 
         # Place trades so open interest is ~400 and target stake is >20000
@@ -160,7 +160,7 @@ if __name__ == "__main__":
         vega.wait_for_total_catchup()
         input(
             "Pause simulation to allow UI to be inspected. "
-            + "Provision from LP_A and LP_B should be deployed."
+            + "Fee should be updated to 0.02 (2.00%)."
         )
 
         # Place trades so open interest is ~600 and target stake is >30000
@@ -176,5 +176,5 @@ if __name__ == "__main__":
         vega.wait_for_total_catchup()
         input(
             "Pause simulation to allow UI to be inspected. "
-            + "Provision from LP_A, LP_B, and LP_C should be deployed."
+            + "Fee should be updated to 0.03 (3.00%)"
         )
--- a/examples/visualisations/multiple_lps.py
+++ b/examples/visualisations/multiple_lps.py
@@ -129,7 +129,6 @@ if __name__ == "__main__":
             vega=vega, market_id=market_id, price=1000, spread=1, volume=10
         )
         vega.wait_fn(120)
-        input("Pause simulation to allow UI to be inspected.")
 
         # Place trades so open interest is ~200 and target stake is >10000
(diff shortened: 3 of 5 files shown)


barnabee:
This scenario looks incorrect. There’s no relationship between commitment amount, target stake for each LP, and whether their liquidity commitment is deployed.

It’s *always* deployed if it can be. Target stake vs commitment amount is only used to choose the fee, the LPs who wanted a higher fee still get their liquidity deployed. They have to manually cancel their commitment if they aren’t happy with the lower fee.

This behaviour is very much intentional.

cdummett:
> This scenario looks incorrect. There’s no relationship between commitment amount, target stake for each LP, and whether their liquidity commitment is deployed.
> 
> It’s _always_ deployed if it can be. Target stake vs commitment amount is only used to choose the fee, the LPs who wanted a higher fee still get their liquidity deployed. They have to manually cancel their commitment if they aren’t happy with the lower fee.
> 
> This behaviour is very much intentional.

Ah thanks @barnabee, I was misunderstanding the spec here, makes sense to deploy all liquidity commitments if possible. 

Originally I understood this as a mechanism to incentivise lower fees (lower fee factor might undercut competitors giving you a larger share of the gathered fees). Is the incentive for an LP submitting a commitment with lower fees purely to attract more traders to a market?

davidsiska-vega:
> Is the incentive for an LP submitting a commitment with lower fees purely to attract more traders to a market?

Basically, yes. 

But for Vega network the incentive is to have as much committed (and deployed) liquidity as possible.

davidsiska-vega (changes requested):
The log messages should be updated to indicate the expected liquidity fees that we should see. 

Also I get an error:

```
Provision from LP_A, LP_B, and LP_C should be deployed.
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/davidsiska/git-repos/github.com/vegaprotocol/vega-market-sim/examples/visualisations/multiple_lps.py", line 181, in <module>
    utils.move_market()
TypeError: move_market() missing 7 required positional arguments: 'vega', 'market_id', 'best_ask_id', 'best_bid_id', 'price', 'spread', and 'volume'

```

barnabee:
> Originally I understood this as a mechanism to incentivise lower fees (lower fee factor might undercut competitors giving you a larger share of the gathered fees). Is the incentive for an LP submitting a commitment with lower fees purely to attract more traders to a market?

Yeah. You can have an incentive to lower fees because lower fees = higher trading volume and that = more fees, so it's like the laffer curve or something — you want to optimise for `fee amount * trading volume` so MMs will be incentivised to do that as trading picks up (and conversely if a market is closed or undersupplied and you *don't* see enough volume to make it worthwhile at the current fee, you could commit liquidity anyway juice the fee up a bit to make up the shortfall).

There are a few reasons you don't want to exclude LPs:

1. more liqudiity is always better for a market and its participants, all else being equal
2. wildly varying liquidity as the supplied vs target ratio keeps changing would be unpredictable for traders and affect slipppage, risk management for LPs who are deployed, etc.
3. as an LP if you're fully excluded (even some of the time) you earn nothing, and i's possinle you could "flicker" in and out of being included, so it's better to keep you in the market earning based on your commitment and let you remove the commitement if you don't want to be there than to leave your funds unproductive

davidsiska-vega (approved):
Works for me now.

PR #291, opened 2023-01-04
feat: rework tags

cdummett:
### Description
Currently the `tag` attribute of an agent is appended to the specified `wallet_name` when creating a wallet for an agent.
As the new scenario setup requires agents to have a unique `name` (`BASE_NAME` + `tag`), the above behaviour results in every agent having a unique wallet which is a bit clunky when debugging using console. Specifying no `tag` results in agents being overwritten when added to the scenario `agents` dictionary.

PR removes above behaviour so the specified `tag` is not appended to the `wallet_name` when creating a wallet.

As part of the PR all existing agents have been standardised to have `key_name` and `tag` arguments (see breaking changes).

### Testing
Passing all tests locally.

### Breaking Changes

Previously for some agent types, specifying a `tag` would create a unique wallet. All agents have been standardised so this behaviour no longer exists. Specifying a `tag` will now only modify the key for the instance of the agent in the `agents` dictionary.

To create a unique wallet or key for an agent, specify a unique `key_name` for the specified `wallet_name`.

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -89,14 +89,15 @@ class MarketOrderTrader(StateAgentWithWallet):
         tag: str = "",
         random_state: Optional[np.random.RandomState] = None,
         base_order_size: float = 1,
-        key_name: str = None,
+        key_name: Optional[str] = None,
         step_bias: Optional[float] = 1,
     ):
-        super().__init__(wallet_name + str(tag), wallet_pass, key_name)
+        super().__init__(
+            wallet_name=wallet_name, wallet_pass=wallet_pass, key_name=key_name, tag=tag
+        )
         self.initial_asset_mint = initial_asset_mint
         self.buy_intensity = buy_intensity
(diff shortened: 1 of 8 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -113,6 +114,10 @@ def initialise(
     ):
         # Initialise wallet
         super().initialise(vega=vega, create_wallet=create_wallet)
+        print(self.wallet_name)
+        print(self.wallet_pass)
+        print(self.key_name)
+        print(self.tag)
```
TomMcL: Don't think we want these 😄

TomMcL (changes requested):
LGTM, just need to remove some prints

vega_sim/scenario/common/agents.py
cdummett: Ah yes thanks! Lazy debugging...

TomMcL (approved):

PR #293, opened 2023-01-05
feat: get_liquidity_fee_shares method

### LINKED ISSUE #286: Add service method for getting / estimating ELS
The `HedgedMarketMaker` agent needs to estimate the fees gained from a trade to accurately set the best-bid and best-ask of its orders when making a market. Currently it assumes it is the only LP in a market to do this.

In the future we may want to add other LPs to the scenario, if this is the case, the `HedgedMarketMaker` will need to be able to either get its exact ELS at the end of an epoch or estimate its current ELS through some means.

cdummett:
### Description
Adds a service method for getting a parties current share of the liquidity fees. New method used in `HedgedMarketMarket` agent to cover the scenario where it is not the only LP.

Also adds fixes to the `HedgedMarketMaker` agent and `HedgedMarket` scenario for changes from #291.

### Examples
```
$ python -m vega_sim.scenario.adhoc -s hedged_market --console --debug
```
### Testing
Passing all tests locally.

### Breaking Changes
None.

### Closes
Closes #286

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -1650,7 +1650,7 @@ class HedgedMarketMaker(ExponentialShapedMarketMaker):
             key_name=self.key_name,
         )
 
-        int_fee = (self.int_mkr_fee + self.int_liq_fee) * fee_share
+        int_fee = self.int_mkr_fee + self.int_liq_fee * fee_share
         ext_fee = self.ext_mkr_fee + self.ext_liq_fee + self.ext_inf_fee
 
         required_bid_price = (
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -1083,6 +1083,9 @@ class ShapedMarketMaker(StateAgentWithWallet):
         self.bid_depth, self.ask_depth = self.best_price_offset_fn(
             current_position, self.current_step
         )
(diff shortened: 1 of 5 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -1638,7 +1641,16 @@ def _optimal_strategy(self, current_position, current_step):
...
+        if (ext_best_bid == 0) or (ext_best_ask == 0):
+            return None, None
+
+        fee_share = self.vega.get_liquidity_fee_shares(
+            market_id=self.market_id,
+            wallet_name=self.wallet_name,
+            key_name=self.key_name,
+        )
+
+        int_fee = (self.int_mkr_fee + self.int_liq_fee) * fee_share
```
TomMcL: Maker fee should still go entirely to them (or not if they aren't the LP who is hit)

vega_sim/service.py
```diff
@@ -2070,3 +2070,23 @@ def list_transfers(
             party_id=party_id,
             direction=direction,
         )
+
+    def get_liquidity_fee_shares(
+        self, market_id: str, wallet_name: str, key_name: Optional[str] = None
+    ) -> Union[Dict, float]:
```
TomMcL, line 2079: This should probably have wallet_name as default `None` too so that it's obvious you can just get with `market_id`

TomMcL (approved):

PR #294, opened 2023-01-05
feat: Specify number of pagination records to avoid datanode issue

TomMcL:
### Description

Specify a number of records to receive from datanode when pagination is required. This works around an issue where if this is unspecified then the `after` flag is ignored and so one ends up in an infinite loop.

### Testing

Tested with code which used to break (RL script) and now completes successfully locally

### Breaking Changes

None, number of records requested is already the default, just needs manually specifying


cdummett (approved):
LGTM - tests and scripts passing locally

PR #295, opened 2023-01-05
chore: Update Vega version

TomMcL:
### Description

Updating Vega core version + minor associated tweaks

### Testing

Tests pass

### Breaking Changes

Should be none


scripts/update_vega_version.sh
```diff
@@ -8,7 +8,7 @@ VEGA_SIM_NETWORKS_TAG=master" > .env
 
 make
 make proto
-pytest -s -v --log-cli-level INFO -m integration
+# pytest -s -v --log-cli-level INFO -m integration
```
cdummett: Is this meant to be commented out still?

TomMcL: It is not

TomMcL: :+1:

cdummett (approved):
LGTM

PR #298, opened 2023-01-10
feat: Removing tag from asset

TomMcL:
### Description

Removing tag from asset in RL runs, meaning that new parties will use the same asset, just on a new market.
This should improve the performance of the RL runs as creating large numbers of assets appears to slow down the simulation run across the day.

### Testing

Tests pass and a few runs of RL look fine. Will run on the nightly Jenkins RL job to see whether it ultimately improves things

### Breaking Changes

None


cdummett (commented):
LGTM - test passing, RL working 👍

cdummett (approved):

davidsiska-vega (approved):

PR #299, opened 2023-01-12
fix: Update function for name change

TomMcL:
### Description

Function name was recently changed, update a usage to use the new name

### Testing

Plots now generate trades feed

### Breaking Changes

None


cdummett (approved):
LGTM

PR #300, opened 2023-01-13
feat: add ParameterExperiment scenario

cdummett:
### Description
PR adds a `ParameterExperiment` scenario for use in both network and market parameter experiments. Scenario has the following key features.

- Configured for 6hrs of trading in 1min steps on an ETHUSD market (gives a 30s sim runtime).
- Can proposes and settles fully configurable markets with the `ConfigurableMarketManager` agent.
- Uses the latest `ExponentialShapedMarketMaker` agent tuned for an ETHUSD market.
- `RandomMarketOrderTraders` agents with un-balanced buy/sell intensities create parties at risk of close-outs.
- Toxic `InformedTrader` agents look ahead and fill orders which will be in the money at a certain point in the future.
- `PriceSensitiveLimitOrderTraders` exploit orders made by the MM when it is overexposed or slow to react to price movements (i.e. when the mm steps after the traders).

OpenInterest and Position
Profit

PR also makes a minor fix to the `InformedTrader` which was not adhering to its configured `max_abs_position` and adds a `PriceSensitiveLimitOrderTrader` which places orders at a random offset from an external price (following an exponential distribution) with the intension of crossing the order book when the market maker is over exposed.

SensitiveTraderProbability

To run the example in console:
```
$ python -m vega_sim.scenario.adhoc -s parameter_experiment --console --pause
```
To run an example parameter test (all parameter tests to be updated with this scenario once merged):
```
$ python -m vega_sim.parameter_test.run -c StakeTargetScaling
```

### Testing
Passing all tests locally.

### Breaking Changes
None

### DIFF
--- a/vega_sim/parameter_test/parameter/configs.py
+++ b/vega_sim/parameter_test/parameter/configs.py
@@ -8,6 +8,7 @@ from vega_sim.parameter_test.parameter.loggers import (
 )
 from vega_sim.scenario.configurable_market.scenario import ConfigurableMarket
 from vega_sim.scenario.curve_market_maker.scenario import CurveMarketMaker
+from vega_sim.scenario.parameter_experiment.scenario import ParameterExperiment
 from vega_sim.scenario.registry import IdealMarketMaker, IdealMarketMakerV2
 from vega_sim.parameter_test.parameter.loggers import (
     BASE_IDEAL_MM_CSV_HEADERS,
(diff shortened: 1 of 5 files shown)


TomMcL (approved):
LGTM

TomMcL:
Looks good. My dream is that one day we can have a scenario setup which takes parameters allowing you to specify "Give me a market like X for Y steps" which would just give a nice somewhat-realistic market acting like some other existing one. I reckon this gets us a bit closer to that too 👍

Issue #304, opened 2023-01-16
Speed up get_trades_from_stream


TomMcL:
`get_trades_from_stream` is doing a copy somewhere which ends up being pretty slow when sim has been running for a while (0.03 seconds). Improve this if possible

PR #310, opened 2023-01-17
feat: trades from stream improvements

TomMcL:
### Description

Adding a filter to trades_from_stream to avoid trades you don't want being included in the return. Also freezing data class such that it can be returned without copying.

Additional various other improvements to streaming and caching, mainly focused around keeping market sim performant under high numbers of repeated trading sessions. RL runs now average 2s out to 100+ iterations. After this point they start to slow, so also added an automatic restart of Vega at this point, which resets the slowdown and allows 2s runs continuously (by this point the training time in current models outweighs the run time significantly, so further improvement is probably not too worthwhile)

image

### Testing

Tests pass

### Breaking Changes

Should be none

### Closes

#304


cdummett (dismissed):
LGTM

davidsiska-vega (changes requested):
Looks good - my only thought is - do we want to hold off till the issue with the RL nightly runs is resolved? 

Or maybe let’s make the restart 100 iterations optional and not do that on the RL nightly runs (at least for now).

davidsiska-vega (dismissed):

vega_sim/service.py
```diff
@@ -154,6 +157,9 @@ def __init__(
         self._aggregated_observation_feed: Queue[Any] = Queue()
         self._kill_thread_sig = threading.Event()
 
+        self._observation_feeds: List[Queue[Any]] = []
+        self._aggregated_observation_feed: Queue[Any] = Queue()
+
```
cdummett: Duplicated?

cdummett (changes requested):
Having an issue when running a scenario I am hanging after settling the market - means all my local tests are hanging as well.

Seems fine in the develop branch.

TomMcL:
@cdummett thanks for catching, looks like the kill signal got lost in the merge! Should be working again now 👍

cdummett (approved):

PR #312, opened 2023-01-18
fix: Fixed thread handling and termination

TomMcL:
### Description

Better handling to ensure threads terminate when they should

### Testing

Locally tests which were hanging now pass

### Breaking Changes

Shouldn't be any


cdummett (approved):
LGTM - all tests passing locally, experiments and scenarios working 👌

PR #316, opened 2023-01-19
feat: add ledger_entries_subscription

cdummett:
### Description
Adds service methods for running a stream listing for ledger entries.

Stream of ledger entries will allow logging of `TRANSFER_TYPE_LIQUIDITY_DISTRIBUTION` events which can be used to calculate the total liquidity fees gained by a party in experiments.

### Testing
Passing all tests locally.

### Breaking Changes
None


vega_sim/service.py
```diff
@@ -140,12 +140,18 @@ def __init__(
         self._market_to_asset = None
         self.seconds_per_block = seconds_per_block
 
+        self.order_thread = None
+        self.transfer_thread = None
+        self.trade_thread = None
```
TomMcL: Looks like these have snuck in

TomMcL (changes requested):

TomMcL:
Slightly concerned about whether we're going to reach a point where Python can't handle all these feeds, but we shall see 😄

cdummett:
> Slightly concerned about whether we're going to reach a point where Python can't handle all these feeds, but we shall see 😄

I agree.. maybe we should go back to optionally turning specific threads on? At the moment only need this thread for tracking liquidity distributions.

vega_sim/service.py
cdummett: Thanks, have removed!

TomMcL (approved):

PR #321, opened 2023-02-02
refactor: add raw call to GetMarketDataHistoryByIDRequest

witgaw:


tests/vega_sim/api/test_data_raw.py
```diff
@@ -1,6 +1,7 @@
 import grpc
 import pytest
 from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
```
TomMcL: Could you change these to just `import datetime`? It's the pattern we've used elsewhere so good to keep consistent :)

TomMcL (commented):
Other than the import change, looks good. Would you mind creating an issue for taking this through `data.py` and adding an interface within `service.py`?

witgaw:
> Other than the import change, looks good. Would you mind creating an issue for taking this through `data.py` and adding an interface within `service.py`?

Sure https://github.com/vegaprotocol/vega-market-sim/issues/325

tests/vega_sim/api/test_data_raw.py
witgaw: done now

TomMcL (approved):
LGTM

PR #326, opened 2023-02-08
feat: devops upgrades

cdummett:
### Description
PR splits `run.py` into `run_scenario.py` and `run_agent.py` which are separate scripts for running whole vega-market-sim scenarios or individual agents on a vega market.

PR also adds the following upgrades to the `VegaServiceNetwork` class and `DevOpsScenario` class: 

- Scripts allow a configurable `step_length_seconds`, the environment will log warning messages if it cannot acheive this step time.
- `MomentumTraders` removed from `DevOpsScenario` to remove `TA-lib` dependency for running network scenarios.
- Better data node switching, `VegaServiceNetwork` will connect/switch to a datanode with an acceptable response-time rather than the first reachable datanode.
- Documentation updated.

### Testing
Passing all changes.

### Breaking Changes
`run.py` removed and replaced with `run_scenario.py` and `run_agent.py`

env variable `VEGA_WALLET_TOKEN` renamed to `VEGA_WALLET_TOKENS_FILE`.

### Closes
None


vega_sim/environment/environment.py
```diff
@@ -190,6 +195,10 @@ def _run(
 
         for agent in self.agents:
             agent.initialise(vega=vega)
+            if isinstance(agent, StateAgentWithWallet):
+                logging.info(
+                    f"{agent.name()} initialised with public_key = {vega.wallet.public_key(name=agent.wallet_name, key_name=agent.key_name)}"
+                )
```
TomMcL, line 215: Why do we need this?

vega_sim/scenario/common/agents.py
```diff
@@ -2773,3 +2773,161 @@ def step(self, vega_state: VegaState):
...
+        open_volume = position.open_volume if position is not None else 0
+
+        if open_volume > 0:
+            self.vega.submit_market_order(
+                trading_wallet=self.wallet_name,
+                key_name=self.key_name,
+                market_id=self.market_id,
+                side="SIDE_SELL",
+                volume=abs(open_volume),
+            )
```
TomMcL: This could be compressed by just checking `if open_volume != 0` then an `X if >0 else Y` on the side

vega_sim/scenario/common/agents.py
```diff
@@ -2773,3 +2773,161 @@ def step(self, vega_state: VegaState):
...
+        self.asset_id = self.vega.find_asset_id(symbol=self.asset_name)
+        if mint_wallet:
+            self.vega.mint(
+                wallet_name=self.wallet_name,
+                asset=self.asset_id,
+                amount=self.initial_asset_mint,
+                key_name=self.key_name,
+            )
+
+        int_market_info = self.vega.market_info(market_id=self.market_id)
```
TomMcL: I'm guessing only one of this and the one on `2835` is needed :)

vega_sim/environment/environment.py
cdummett: It's a work around for not being able to connect to a wallet in console at the moment.

We can use the "View As Vega User" feature with one of the keys logged here to access all their info.

vega_sim/network_service.py
```diff
@@ -387,9 +414,7 @@ def data_node_query_url(self) -> str:
     @property
     def wallet_url(self) -> str:
         if self._wallet_url is None:
-            self._wallet_url = (
-                f"http://{self.network_config['Host']}:{self.network_config['Port']}"
-            )
+            self._wallet_url = f"http://127.0.0.1:1789"
```
TomMcL, line 417: Why this change?

vega_sim/wallet/vega_wallet.py
```diff
@@ -216,7 +217,12 @@ def get_keypairs(self, wallet_name: str) -> dict:
         url = f"{self.wallet_url}/api/v2/requests"
 
         response = requests.post(url, headers=headers, json=submission)
-        response.raise_for_status()
+
+        try:
+            response.raise_for_status()
+        except Exception as e:
+            logging.warning(f"Request failed, response={response.json()}")
```
TomMcL, line 224: In principle this should probably be more than a warning, but works for now :)

vega_sim/network_service.py
cdummett: So not all our network config files have a host and port specified e.g. [testnet2](https://github.com/vegaprotocol/networks/blob/master/testnet2/testnet2.toml).

But I think this is the default url console expects across all networks so feels safe setting this wallet url?

TomMcL: Yeah we should probably move to using the generic config location wallet uses now, but I guess this is a reasonable stopgap!

TomMcL (approved):

PR #335, opened 2023-02-16
feat: Environment and scenario for 'realtime' market sim

TomMcL:
### Description

Environment and fauceting agent for a scenario which first sets up a market/set of markets with specified asset and faucets parties an amount of each asset then ticks blocks at a specified rate, allowing an environment which progresses forwards in time at a more regular rate for interaction with external bots.

### Testing

Tested using external bots coded in Java which interact with a market at "real world" speed.

### Breaking Changes

Should be none


cdummett (approved):
LGTM

PR #344, opened 2023-02-24
feat: Functionality to automatically download binaries when missing

TomMcL:
### Description

Adding a check which will download binaries if they are missing when trying to start up nullchain

### Testing

Locally tested with `bin` folder deleted + as a package

### Breaking Changes

Should be none, if the `bin` folder exists it will be used


cdummett (approved):
LGTM

PR #345, opened 2023-02-28
feat: update MM to use GTT orders

cdummett:
### Description
Updates MM to use `GTT` orders where the `expiry_at` field is set as some multiple of the agents `step_length_seconds`. If no `step_length_seconds` is specified the agent defaults too using `GTC` orders.

### Testing
Passing all tests locally.

### Breaking Changes
None as MM will use `GTC` if no `step_length_seconds` specified.


vega_sim/scenario/common/agents.py
```diff
@@ -1315,6 +1303,12 @@ def _move_side(
         submissions = []
         cancellations = []
 
+        expires_at = (
+            int((self.vega.get_blockchain_time() + 5 * self.step_length_seconds) * 1e09)
```
TomMcL:
```suggestion
            int((self.vega.get_blockchain_time() + 5 * self.step_length_seconds) * 1e9)
```

TomMcL (changes requested):
The target of the PR sounds good to me, though I'd probably rename `step_length_seconds` to something more descriptive about what it's target is (if we want to default it to `step_length_seconds` when building the agent then that's fine). And why the `*5`?

vega_sim/scenario/common/agents.py
cdummett: `* 5 ` was a safety margin, sometimes the environment may take longer to step then an agents `step_length_seconds` e.g. has to switch data-node or a query is slow). I'll make this configurable.

I named it `step_length_seconds` as I wanted it to be clear it was the same values as the `step_length_seconds` of the environment. Not sure how I would default the value.

Maybe it should be an agent wide property set in the `initialise` method?

TomMcL: Could we not achieve the same thing slightly more clearly, and allow the agent itself to be more flexible, with having the parameter `order_validity_seconds` and setting that to `step_length_seconds * 5` when that's what we want?

cdummett: Ah nice yeah this is much clearer, will make that chage 👍

TomMcL (approved):

PR #348, opened 2023-03-09
feat: fix up local wallet nullchain

TomMcL:
### Description

Fixing up function calls so that null chain works with a full local wallet once more

### Testing

Tested with running full local wallet

### Breaking Changes

Should be none


cdummett (approved):
LGTM - just noticed I don't think we have any integration tests using the full vega wallet. May be worth adding in a future PR?

TomMcL:
@cdummett Yes agreed, can just duplicate one or some of the existing tests with a full wallet and check they pass too

PR #359, opened 2023-03-20
feat: fuzz test plots

### LINKED ISSUE #352: Metrics / Stats / Visualisations for the fuzzing run
It would be grand to have metrics produced by the fuzzing run - https://github.com/vegaprotocol/vega-market-sim/issues/346  (that ideally get spat out into the slack channel but that's not essential). 

What I'd be looking for:
- time elapsed (how long did we pretend the chain ran). Was it days? Weeks?
- number of trades placed, volume, traded
- mark price volatility vs. vol of the "price source" we're following
- number of closeout trades / closeout volume
- number of times LP got closed out 
- fraction of time in various states (price monitoring / liquidity monitoring)... ideally we'd cook things up so that these things are hit but we're not spending too much time stuck in auctions (but any improvements can be round 2, round 1 is get the metrics). 
- various parties P&L (e.g. the non-suicidal LPs should be making money)
- @TomMcL / @cdummett anything else interesting you can think of. 
- @TomMcL's excellent market visualisation that's produced for the RL runs. 

As a bonus:
- collect these across successive runs - we can start to get a feeling for which ones are "stable" and which ones are "unstable". Eventually we should monitor the stable ones and justify any jumps (in terms of desired changes in core).

cdummett:
### Description

PR adds a script for running the fuzz test as well as methods for creating figures for summarising test data.

#### Fuzz Testing Summary Plot

- **Trading Mode Analysis**: figure showing the current trading mode (upper plot) and the proportion of time spent in each trading mode (lower plot).
- **Price Analysis**: figure showing the external price and mark price over time. Figure also compares the volatility of the external-price and the volatility of the mark-price over the simulation period.
- **Close Out Analysis**: figure showing the cumulative number of trader close-outs and LP close-outs compared to the funds in the insurance pool.

fuzz-d74feeb3a320b884f18f13c8a3a9aef85969c02e831751b6a2bd1ee80d1cdfe0

#### Fuzz Trader Coverage Plot
newplot (1)

### Testing
Passing all tests locally.

To test manually run a short fuzz test, plots should be generated:
```
python -m vega_sim.scenario.fuzzed_markets.run_fuzz_test --steps 100
```

### Breaking Changes
None

### Closes
Closes #352


davidsiska-vega:
Question:
We will have two of these each, assuming we're doing 2 markets? Or perhaps two of the "fuzz testing plots" and one overall "fuzz trader coverage"?

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -123,17 +140,33 @@ def create_fuzzed_amendment(self):
...
-            side=self.random_state.choice(
-                a=["SIDE_UNSPECIFIED", "SIDE_BUY", "SIDE_SELL"],
-            ),
-            size=self.random_state.poisson(lam=10),
-            order_type=self.random_state.choice(
-                a=["TYPE_UNSPECIFIED", "TYPE_MARKET", "TYPE_LIMIT"]
-            ),
-            time_in_force=self.random_state.choice(
+    def create_fuzzed_submission(self, vega_state):
+        self.__class__.MEMORY["TRADING_MODE"].append(
```
TomMcL: You actually shouldn't need this class funkiness here (and it's a 'feature' of which one needs to be aware, but is useful in this case!) the `MEMORY` dict is shared in any case as it is defined at top level:

image

cdummett:
> Question: We will have two of these each, assuming we're doing 2 markets? Or perhaps two of the "fuzz testing plots" and one overall "fuzz trader coverage"?

One fuzzing plot and one of Toms trading plots for each market and then a single fuzzing coverage plot aggregated for all the markets @davidsiska-vega .

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -173,6 +218,38 @@ def _select_order_id(self):
         else:
             return None
 
+    def finalise(self):
+        if not self.__class__.OUTPUTTED:
```
TomMcL: Can we add an `output_plot_on_finalise` flag to the class then bring the plotly import actually inside this function? Would avoid having to have a top level import for plotting stuff when people may not always want/need

vega_sim/scenario/fuzzed_markets/agents.py
cdummett: So this is exactly how I thought it should work originally but it doesn't seem to... not sure if I am missing something painfully obvious but removing the `__class__` results in a unique `MEMORY` for each market and a unique `OUTPUTTED` flag for each agent! So when the environment is finalised I get 20 figures created, 10 of Market A and 10 of Market B! Any ideas?

cdummett: @TomMcL all sorted, I was redefining the class variable rather than modifying 🤦 
```
class MyClass:
    class_var = 10

    def modify_class_var(self, value):
        MyClass.class_var += value

    def redefine_class_var(self, value):
        self.class_var = value

if __name__ == "__main__":

    obj1 = MyClass()
    obj2 = MyClass()

    # modifying the class variable
    obj1.modify_class_var(5)
    print(obj1.class_var)     # Output: 15
    print(obj2.class_var)     # Output: 15

    # redefining the class variable
    obj1.redefine_class_var(20)  
    print(obj1.class_var)     # Output: 20
    print(obj2.class_var)     # Output: 15
```

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -192,10 +193,10 @@ def create_fuzzed_submission(self, vega_state):
 
         return self.vega.create_order_submission(
             market_id=self.market_id,
-            side=self.__class__.MEMORY["SIDE"][-1],
+            side=FuzzingAgent.MEMORY["SIDE"][-1],
```
TomMcL, line 196: `self` works here too but perhaps this is clearer anyway!

TomMcL (approved):
LGTM

I wonder whether there's any additional way (in a later change) to show the top test cases which are missing. Slightly tricky as you'd have to have a list of what you expected to see, but currently anything which hasn't been tested gets smaller and so harder to see 😄

cdummett:
> LGTM
> 
> I wonder whether there's any additional way (in a later change) to show the top test cases which are missing. Slightly tricky as you'd have to have a list of what you expected to see, but currently anything which hasn't been tested gets smaller and so harder to see 😄

This is a very good point, maybe they should be bigger if they are covered less, basing the size of `1/count` would be a quick way to do this. I have linked your comment to this [issue](https://github.com/vegaprotocol/vega-market-sim/issues/363) for the second round of improvements.

PR #362, opened 2023-03-21
feat: Update vega version + needed changes to go with

TomMcL:
### Description

Updating Vega core version plus associated changes. Systematisation of where we use nanoseconds vs seconds and passing a live flag to liquidity provisions as this was added.

### Testing

Tests were broken and now pass

### Breaking Changes

`get_blockchain_time` which used to default to seconds now defaults to nanoseconds. All cases of this use should have been updated


cdummett (dismissed):
LGTM

cdummett (approved):
LGTMv2

PR #366, opened 2023-03-22
fix: No longer need 1e9 multiple

TomMcL:
### Description

Removing multiple from when timestamps were seconds

### Testing

Integration tests only


cdummett (approved):
LGTM

PR #371, opened 2023-03-23
feat: Update version of Vega

TomMcL:
### Description

Updating vega core version plus requisite changes

### Testing

Integration tests

### Breaking Changes

Should be none


vega_sim/api/data_raw.py
```diff
@@ -272,15 +272,19 @@ def list_orders(
...
+
+    if reference is not None:
+        order_filter.reference = reference
 
     for attr, val in [
-        ("market_id", market_id),
-        ("party_id", party_id),
-        ("reference", reference),
+        ("market_ids", [market_id] if market_id is not None else None),
+        ("party_ids", [party_id] if party_id is not None else None),
```
cdummett, line 282: Doesn't have to be in this PR but at some point I guess we could update the methods to allow a lists of ids? Would condense the `initialise_order_monitoring` method in the `local_data_cache` a fair bit.

cdummett (approved):
LGTM

vega_sim/api/data_raw.py
TomMcL: 👍

PR #372, opened 2023-03-23
feat: switch decimal cache to use feed

cdummett:
### Description
PR adds listeners for asset and market creation events and switches the `DecimalCache` factory methods to use the `local_data_cache` "from feed" methods rather than data-node queries.

Should resolve the issue where market data bus events are observed before the datanode has processed a market creation event.

PR also removes some waits which are no longer needed.

### Testing
Passing all tests locally (not yet).

### Breaking Changes
None


vega_sim/local_data_cache.py
```diff
@@ -283,6 +309,22 @@ def start_live_feeds(
...
+                self._asset_from_feed[asset.id] = asset
+
+    def initialise_markets(self):
+        base_markets = [
+            market.id for market in data_raw.all_markets(self._trading_data_client)
+        ]
+
+        with self.market_lock:
+            for market in base_markets:
+                self._market_from_feed[market.id] = market
```
TomMcL, line 326: A lot of these (outside this change) we could quite possibly remove locks from, anything we only update/read and never iterate over to read should be fine to drop a lock from. Not for now though

vega_sim/local_data_cache.py
```diff
@@ -419,6 +461,24 @@ def _monitor_stream(self) -> None:
                 elif isinstance(update, events_protos.TimeUpdate):
                     with self.time_update_lock:
                         self._time_update_from_feed = int(update.timestamp)
+
+                elif isinstance(update, vega_protos.markets.Market):
+                    with self.market_data_lock:
+                        self._market_from_feed[update.id] = update
+                        # Trigger save by attempting to get decimal for market
```
TomMcL: Doesn't need to do these within the lock

vega_sim/local_data_cache.py
```diff
@@ -419,6 +461,24 @@ def _monitor_stream(self) -> None:
...
+                        self._market_pos_decimals[update.id]
+                        self._market_price_decimals[update.id]
+                        self._market_to_asset[update.id]
+
+                elif isinstance(update, vega_protos.assets.Asset):
+                    with self.trades_lock:
+                        self._asset_from_feed[update.id] = update
+                        # Trigger save by attempting to get decimal for asset
```
TomMcL: ditto

TomMcL (changes requested):

TomMcL (dismissed):

TomMcL (approved):

PR #379, opened 2023-03-28
feat: Add gRPC readiness check to criteria on startup

TomMcL:
### Description

Adding a check for gRPC readiness specifically as it seems REST was returning before gRPC was returning valid values from datanode

### Testing

Tried a couple of tests, all worked fine, will observe output of full run.


cdummett (approved):
lgtm

PR #380, opened 2023-03-28
fix: missing builtin assets on startup

cdummett:
### Description
None

### Testing
None

### Breaking Changes
None

### Closes
None


TomMcL (approved):
Sure why not

PR #386, opened 2023-04-02
feat: add account balance plots

cdummett:
### Description

PR adds functionality to the `Scenario` to get the dictionary of agents stored in the `Snitch` as well as methods for storing agent information in a `.csv` post run.

The `.csv` file maps agent keys to agent types, tags, and names.

PR also adds tools for plotting account history by agent type.
accounts-c71e76cf9b8ee91e265c0d2387f3652abc6ae15d5ffeec731ca9a32bed0eda8a

### DIFF
--- a/vega_sim/scenario/scenario.py
+++ b/vega_sim/scenario/scenario.py
@@ -6,7 +6,10 @@ from vega_sim.null_service import VegaService
 from vega_sim.environment.environment import MarketEnvironment
 from vega_sim.scenario.constants import Network
 from vega_sim.scenario.common.agents import Snitch, StateAgent, MarketHistoryData, Agent
-from vega_sim.tools.scenario_output import market_data_standard_output
+from vega_sim.tools.scenario_output import (
+    market_data_standard_output,
+    agents_standard_output,
+)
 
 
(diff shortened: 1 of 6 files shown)


vega_sim/scenario/scenario.py
```diff
@@ -97,6 +101,10 @@ def run_iteration(
     def get_snitch(self) -> Optional[Snitch]:
         return self.agents.get("snitch")
 
+    def get_agents(self) -> dict:
+        snitch = self.get_snitch()
+        return snitch.agents if snitch is not None else {}
```
TomMcL: This seems slightly confusing that it defaults to an empty dictionary? Why not `self.agents` (and what is the reason to use this rather than just `self.agents`? I can't see it used in the PR

cdummett: Ah thanks this is left over from an earlier approach!

 The scenario `self.agents` dict contains the `Snitch` whilst the snitches `self.agents` dict does not. In the earlier approach having that Snitch was causing a problem but this whole method is now redundant.

Will removed it 👍

TomMcL (approved):

PR #395, opened 2023-04-20
feat: add price variation to fuzz test

cdummett:
### Summary

PR improves the fuzz test to better test price-monitoring auctions and loss-socialisation.

### Description

The `price_process` used in the fuzz test has been modified to include the following:
- price-drifts: periods where the price-process drifts
- price-spikes: temporary spikes in the price-process which are corrected

The price-spikes should trigger multiple triggers simultaneously, but only result in a short price-monitoring auction as the market corrects itself. The price-drifts however could trigger multiple triggers at the end of an auction resulting in auction extensions.

To better visualise price-monitoring auctions a `price_monitoring_plot` method has been added:

Long run example:
monitoring-1

Short run example:
monitoring-0

As a result of the above changes, markets can now exit price-monitoring auctions at a distant price. This causes larger price differences between mark-to-markets which can encourage using funds from the insurance pool and loss-socialisation (degenerate traders can't cover their losses).

fuzz-1

### Testing
Passing all tests locally.

### Breaking Changes
None

### Closes
None

### DIFF
--- a/vega_sim/api/data.py
+++ b/vega_sim/api/data.py
@@ -510,10 +510,10 @@ def _market_data_from_proto(
         auction_end=market_data.auction_end,
         auction_start=market_data.auction_start,
         indicative_price=num_from_padded_int(
-            market_data.static_mid_price, decimal_spec.price_decimals
+            market_data.indicative_price, decimal_spec.price_decimals
         ),
         indicative_volume=num_from_padded_int(
-            market_data.static_mid_price, decimal_spec.price_decimals
+            market_data.indicative_volume, decimal_spec.price_decimals
         ),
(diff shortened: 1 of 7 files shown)


davidsiska-vega (approved):
I am assuming the price monitoring plots will go into the artifacts place on overnight runs. Looking forward to it!

cdummett:
> I am assuming the price monitoring plots will go into the artifacts place on overnight runs. Looking forward to it!

@davidsiska-vega Yeah price monitoring plots have been added to the artifacts - can see them in the current build already https://jenkins.ops.vega.xyz/job/common/job/vega-market-sim-reinforcement/173/

PR #405, opened 2023-05-03
feat: add estimate position method and tests

cdummett:
### Description

To allow market-sim to test the new `EstimatePosition` data-node endpoint, api methods and service methods needed to be added to market-sim.

PR adds the following methods each with a suitable test: 

- `estimate_position` method added to the `data_raw` module: 
    - handles sending a request to the endpoint and returning a raw response
- `estimate_position` method added to the `data` module:
    - handles converting the raw response into equivalent market-sim data classes
- `estimate_position` method added to the service module:
    - handles interface between market-sim and api methods.

### Testing
Passing all tests locally.

### Breaking Changes
None


Jiajia-Cui (approved):
LGTM

PR #409, opened 2023-05-05
feat: add test for estimatePosition

Jiajia-Cui:
closes [issue #2076](https://github.com/vegaprotocol/system-tests/issues/2076)

### DIFF
--- a/tests/integration/test_trading.py
+++ b/tests/integration/test_trading.py
@@ -2,6 +2,7 @@ import pytest
 import logging
 from collections import namedtuple
 import vega_sim.proto.vega as vega_protos
+from examples.visualisations.utils import continuous_market,move_market
 
 from tests.integration.utils.fixtures import (
     ASSET_NAME,
@@ -298,19 +299,19 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
     )
 
 @pytest.mark.integration
-def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
+def test_liquidation_and_estimate_position_calculation(vega_service: VegaServiceNull):
     vega = vega_service
     WalletConfig = namedtuple("WalletConfig", ["name", "passphrase"])
 
     MM_WALLET = WalletConfig("mm", "pin")
     PARTY_A = WalletConfig("party_a", "party_a")
     PARTY_B = WalletConfig("party_b", "party_b")
+    PARTY_C = WalletConfig("party_c", "party_c")
     TERMINATE_WALLET = WalletConfig("TERMINATE", "TERMINATE")
     LIQ_WALLET = WalletConfig("LIQ", "TRADER")
     ASSET_NAME = "tDAI"
-    WALLETS = [MM_WALLET, PARTY_A, PARTY_B, TERMINATE_WALLET, LIQ_WALLET]
+    WALLETS = [MM_WALLET, PARTY_B, PARTY_C, TERMINATE_WALLET, LIQ_WALLET]
 
-    # vega: VegaServiceNull,
     mint_amount = 10000
     initial_volume = 10
     initial_commitment = 1000
@@ -319,6 +320,7 @@ def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
     vega.wait_for_total_catchup()
     for wallet in WALLETS:
         vega.create_key(wallet.name)
+    vega.create_key(PARTY_A.name)
 
     vega.wait_for_total_catchup()
     vega.mint(
@@ -328,7 +330,6 @@ def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
     )
     vega.forward("10s")
 
-    # Create asset
     vega.create_asset(
         MM_WALLET.name,
         name=ASSET_NAME,
@@ -347,6 +348,12 @@ def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
             asset=asset_id,
             amount=mint_amount,
         )
+
+    vega.mint(
+        PARTY_A.name,
+        asset=asset_id,
+        amount=619,
+    )
     vega.forward("10s")
     vega.create_simple_market(
         market_name="CRYPTO:BTCDAI/DEC22",
@@ -356,6 +363,9 @@ def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
         termination_key=TERMINATE_WALLET.name,
         market_decimals=0,
     )
+    vega.update_network_parameter(
+            MM_WALLET.name, parameter="network.markPriceUpdateMaximumFrequency", new_value="0"
+        )
 
     market_id = vega.all_markets()[0].id
 
@@ -364,17 +374,18 @@ def test_liquidation_and_estimate_position(vega_service: VegaServiceNull):
         market_id=market_id,
         commitment_amount=initial_commitment,
         fee=0.002,
-        buy_specs=[("PEGGED_REFERENCE_MID", 50, 1)],
-        sell_specs=[("PEGGED_REFERENCE_MID",50, 1)],
+        buy_specs=[("PEGGED_REFERENCE_BEST_BID", 50, 1)],
+        sell_specs=[("PEGGED_REFERENCE_BEST_ASK",50, 1)],
         is_amendment=False,
     )
     # Add transactions in the proposed market to pass opening auction at price 1000
     vega.submit_order(
-        trading_key=PARTY_A.name,
+        trading_key=PARTY_C.name,
         market_id=market_id,
         order_type="TYPE_LIMIT",
         time_in_force="TIME_IN_FORCE_GTC",
         side="SIDE_BUY",
+        order_id="best_bid_id",
         price=800,
         volume=initial_volume,
     )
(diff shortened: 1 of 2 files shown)


tests/integration/test_trading.py
```diff
@@ -294,3 +296,175 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
         is_market_order=[False, False],
         collateral_available=1,
     )
+
+@pytest.mark.integration
+def test_liquidation_and_estimate_position(vega_service_with_market: VegaServiceNull):
```
cdummett: The `vega_service_with_market` fixture is going to create a market and as you are creating your own market with the same asset symbol and market name as the fixture it could cause some confusion in which one the test is actually using.

When creating a market within the test it may be better to use the `vega_service` fixture.

tests/integration/test_trading.py
```diff
@@ -294,3 +296,175 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        MM_WALLET.name,
+        name=ASSET_NAME,
+        symbol=ASSET_NAME,
+        decimals=1,
+        max_faucet_amount=10 * mint_amount * 1e5,
+    )
+    vega.forward("10s")
+    vega.wait_for_total_catchup()
+
+    asset_id = vega.find_asset_id(symbol=ASSET_NAME)
```
cdummett, line 471: The current fixture `vega_service_with_market` also creates the `tDAI` asset so this method may not return the asset id you are after. Try switching the fixture so there is only one asset on the network.

tests/integration/test_trading.py
```diff
@@ -294,3 +296,175 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    vega.forward("10s")
+    vega.create_simple_market(
+        market_name="CRYPTO:BTCDAI/DEC22",
+        proposal_key=MM_WALLET.name,
+        settlement_asset_id=asset_id,
+        termination_key=TERMINATE_WALLET.name,
+        market_decimals=1,
+    )
+
+    market_id = vega.all_markets()[0].id
```
cdummett, line 486: The current fixture `vega_service_with_market` also creates a market so this method may not return the market id you are after. Try switching the fixture so there is only one market on then network.

For extra safety use the `find_market_id` method instead and pass the market name.

tests/integration/test_trading.py
```diff
@@ -294,3 +296,175 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    linear_slippage_factor = float(market_info.linear_slippage_factor)
+    quadratic_slippage_factor = float(market_info.quadratic_slippage_factor)
+    market_data = vega.get_latest_market_data(market_id=market_id)
+    markprice = market_data.mark_price
+
+    estimate_margin_open_vol_only, estimate_liquidation_price_open_vol_only = vega.estimate_position(
+        market_id,
+        open_volume=1,
+        side=["SIDE_SELL", "SIDE_SELL"],
+        price=[1.01, 1.02],
```
cdummett: These prices, sell prices well below the mark price, cause some unexpected behaviour for me. As with the `collateral_available` comment maybe start with the happy case before edge cases.

tests/integration/test_trading.py
```diff
@@ -294,3 +296,175 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    TERMINATE_WALLET = WalletConfig("TERMINATE", "TERMINATE")
+    LIQ_WALLET = WalletConfig("LIQ", "TRADER")
+    ASSET_NAME = "tDAI"
+    WALLETS = [MM_WALLET, PARTY_A, PARTY_B, TERMINATE_WALLET, LIQ_WALLET]
+
+    # vega: VegaServiceNull,
+    mint_amount = 10000
+    initial_volume = 10
+    initial_commitment = 1000
+    collateral_available = 1
```
cdummett: This low `collateral_available` value is causing some unexpected behaviour, i.e. `liquidation_price` above the `mark_price` for a long position. Maybe start by checking the happy case before looking at edge cases.

cdummett (commented):
I know only a draft but some suggestions I noticed whilst debugging 😄

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
     )
+
+@pytest.mark.integration
+def test_liquidation_and_estimate_position_calculation_AC001003(vega_service: VegaServiceNull):
+    vega = vega_service
+    WalletConfig = namedtuple("WalletConfig", ["name", "passphrase"])
+
+    MM_WALLET = WalletConfig("mm", "pin")
+    PARTY_A = WalletConfig("party_a", "party_a")
+    PARTY_B = WalletConfig("party_b", "party_b")
```
cdummett: These are going to conflict with the module level variables `PARTY_A` and `PARTY_B` defined on line 21 and 22. Let's either use the module level variables in the functions or create new function level variables with lower case.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    PARTY_C = WalletConfig("party_c", "party_c")
+    TERMINATE_WALLET = WalletConfig("TERMINATE", "TERMINATE")
+    LIQ_WALLET = WalletConfig("LIQ", "TRADER")
+    ASSET_NAME = "tDAI"
+    WALLETS = [MM_WALLET, PARTY_B, PARTY_C, TERMINATE_WALLET, LIQ_WALLET]
```
cdummett: These should also be lower-case as they are only within the scope of the function.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+            amount=mint_amount,
+        )
+    vega.mint(
+        PARTY_A.name,
+        asset=asset_id,
+        amount=collateral_available,
+    )
+    # Wait for the market creation to complete
+    marketManager.vega.wait_for_total_catchup()
+    market_id = vega.all_markets()[0].id
```
cdummett, line 486: Let's use `vega.find_market_id` here instead. That method has some error handling built-in in case the market creation fails.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+
+    # Access the updated value
+    print(f"linear slippage factor = {configWithSlippage.linear_slippage_factor}")
```
cdummett: Best not to have prints in the final code, we could either remove them entirely or if you think they are still useful, we can use logging instead here.
```python
logging.debug(f"linear slippage factor = {configWithSlippage.linear_slippage_factor}")
```

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        volume=initial_volume,
+    )
+    #AC 0012-NP-LIPE-003: An estimate is obtained for a position with no open volume and a single limit buy order, after the order fills the mark price keeps going down in small increments and the actual liquidation takes place within the obtained estimated range. 
+    estimate_margin_open_vol_only_before_trade, estimate_liquidation_price_including_buy_orders = vega.estimate_position(
+        market_id,
+        open_volume=0,
+        side=["SIDE_BUY"],
+        price=[1000],
+        remaining=[10],
+        is_market_order=[False, False],
```
cdummett: The length of the `side`, `price`, `remaining` and `is_market_order` fields should be the same. Let's drop that second `False` from `is_market_order`.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        vega.create_key(wallet.name)
+    vega.create_key(PARTY_A.name)
+
+    vega.wait_for_total_catchup()
+    vega.mint(
+        MM_WALLET.name,
+        asset="VOTE",
+        amount=1e4,
+    )
+    vega.forward("10s")
```
cdummett: Don't think we need this line, we fast forward time below anyway.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    vega.wait_for_total_catchup()
+    vega.wait_fn(1)
```
cdummett: General point, make sure we `wait_for_total_catchup` after we fast forward time, other wise datanode can be out of sync.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
         is_market_order=[False, False],
         collateral_available=1,
     )
+
+@pytest.mark.integration
+def test_liquidation_and_estimate_position_calculation_AC001003(vega_service: VegaServiceNull):
+    vega = vega_service
+    WalletConfig = namedtuple("WalletConfig", ["name", "passphrase"])
```
cdummett: `WalletConfig` is already imported in line 11 so no need to define it again here.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        MM_WALLET.name,
+        asset="VOTE",
+        amount=1e4,
+    )
+    vega.forward("10s")
+    vega.wait_for_total_catchup()
+    vega.wait_fn(1)
+
+    configWithSlippage = MarketConfig()
+    configWithSlippage.load()  # Load the default configuration
```
cdummett: Don't need this line, a `MarketConfig` instance will call `self.load()` within its `__init__` method.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    configWithSlippage.set("linear_slippage_factor", str(1))  # Set the linear_slippage_factor to 0.5
+    configWithSlippage.set("quadratic_slippage_factor", str(1))  # Set the quadratic_slippage_factor to 0.5
+    configWithSlippage.set("decimal_places", int(0))  # Set the market decimal_places to 0
+    configWithSlippage.set("lp_price_range", str(1))
+
+    # Initialize the Market Manager
+    marketManager = ConfigurableMarketManager(
+        proposal_key_name="MM_WALLET.name",
```
cdummett: This works but just to be clear this is setting the key to the string `"MM_WALLET.name"` rather than the value of `MM_WALLET.name` which is `"mm"`. Might be better off using `MM_WALLET.name`.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        trading_key=PARTY_A.name,
+        market_id=market_id,
+        order_type="TYPE_LIMIT",
+        time_in_force="TIME_IN_FORCE_GTC",
+        side="SIDE_BUY",
+        price=1000,
+        volume=initial_volume,
+    )
+    #AC 0012-NP-LIPE-003: An estimate is obtained for a position with no open volume and a single limit buy order, after the order fills the mark price keeps going down in small increments and the actual liquidation takes place within the obtained estimated range. 
+    estimate_margin_open_vol_only_before_trade, estimate_liquidation_price_including_buy_orders = vega.estimate_position(
```
cdummett: We can "drop" the `estimate_margin_open_vol_only_before_trade` variable by replacing it with `_` if we don't need it.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+    linear_slippage_factor = float(market_info.linear_slippage_factor)
+    quadratic_slippage_factor = float(market_info.quadratic_slippage_factor)
+  
+    estimate_margin_open_vol_only, estimate_liquidation_price_open_vol_only = vega.estimate_position(
+        market_id,
+        open_volume=10,
+        side=["SIDE_BUY"],
+        price=[1000],
+        remaining=[0],
+        is_market_order=[False, False],
```
cdummett: This order doesn't exist on the book any more and instead the party has `open_volume`, lets remove these fields.

tests/integration/test_trading.py
```diff
@@ -294,3 +299,795 @@ def test_estimate_position(vega_service_with_market: VegaServiceNull):
...
+        volume=initial_volume,
+    )
+    vega.wait_fn(1)
+    vega.wait_for_total_catchup()
+    #AC 0012-NP-LIPE-001:Check actual liquidation price is within the estimated price 
+    print(f"market after amending at new price {vega.get_latest_market_data(market_id=market_id)}")
+    print(f"account_PARTY_A is closeout since sum of account = {vega.party_account(key_name=PARTY_A.name, asset_id=asset_id, market_id=market_id).general+vega.party_account(key_name=PARTY_A.name, asset_id=asset_id, market_id=market_id).margin}")
+    assert vega.party_account(key_name=PARTY_A.name, asset_id=asset_id, market_id=market_id).general+vega.party_account(key_name=PARTY_A.name, asset_id=asset_id, market_id=market_id).margin ==0
+    assert estimate_liquidation_price_open_vol_only.best_case.open_volume_only >= vega.get_latest_market_data(market_id=market_id).mark_price
+    assert estimate_liquidation_price_open_vol_only.worst_case.open_volume_only<= vega.get_latest_market_data(market_id=market_id).mark_price
```
cdummett, line 648: Looks like we are only checking the mark price relative to `estimate_liquidiation_price_open_vol_only`, i.e. the estimate from when the trader had a position (covering of `AC-NP-LIPE-001`).

Do we not also need to check relative to `estimate_liquidiation_price_including_buy_orders` to validate `AC-NP-LIPE-003`.

cdummett, line 648: There are a couple of places here where we can reduce the number of data-node queries we use by making the call once and then using the response in each of the assertions.

cdummett (changes requested):
Looks good, have only looked at the first test so far and added some small points.

I think it could be improved slightly if we add a few more price movements (I think we need at least two mark price movements) e.g. one to a price just before the expected worst-case liquidation price (the party is not closed out) and a second to a price just after the expected best-case liquidation price (the party is definitely closed out).

At the moment the true liquidation price could be anywhere in the range `1000` -> `985`?

Jiajia-Cui:
close and replace it with https://github.com/vegaprotocol/vega-market-sim/pull/425

PR #411, opened 2023-05-10
feat: fuzzing runs on vegacapsule network

### LINKED ISSUE #406: Add Network History Segment comparison to market sims tests
To try and capture non-deterministic behaviour across core/datanode we would like to compare the network segments produced by nightly market sims test. The test should compare the segments produced by 2 datanodes connected to seperated core nodes.

cdummett:
### Description
PR adds functionality to launch a vega network with vegacapsule and then run a scenario against it.

### Local Instructions

1. run `make capsule`
2. run `sh scripts/build-docker-learning.sh`
3. run `sh scripts/run-capsule-test.sh 100`

close #406


vega_sim/environment/environment.py
```diff
@@ -518,11 +534,11 @@ def step(self, vega: VegaServiceNetwork) -> None:
             try:
                 agent.step(state)
             except Exception as e:
-                msg = f"Agent '{agent.key_name}' failed to step. Error: {e}"
+                msg = f"Agent '{agent.name()}' failed to step. Error: {e}"
                 if self.raise_step_errors:
                     raise e(msg)
                 else:
-                    logging.warning(msg)
+                    logging.debug(msg)
```
TomMcL: I think these should probably still be warnings (even if it does fill up the log occasionally with faucet errors) as it's useful to see and I worry we'll forget that it exists otherwise

vega_sim/local_data_cache.py
```diff
@@ -196,7 +199,10 @@ def asset_from_feed(
     ) -> vega_protos.assets.Asset:
         if asset_id not in self._asset_from_feed:
             self.initialise_assets()
-        return self._asset_from_feed[asset_id]
+        try:
+            return self._asset_from_feed[asset_id]
+        except KeyError as e:
+            raise e
```
TomMcL: Am I missing something here, what does this do?

cdummett: Ah yes good catch, left over from when I was excepting errors when developing (the raise was a log). Will remove this 👍

TomMcL (approved):

PR #413, opened 2023-05-17
feat: tie to mainnet version

TomMcL:
### Description

Setting version of Vega used to be the same commit currently used for main net such that we can generate a package on that version

### Testing

Running tests

### Breaking Changes

None

### Closes

None


.env
```diff
@@ -1,4 +1,4 @@
-VEGA_SIM_VEGA_TAG=ae553695aab6ebf72463d4d626bd4bfc3fbd123c
+VEGA_SIM_VEGA_TAG=61d1f77ee360bf1679d5eb0e0efdb1cce977c9db
```
davidsiska-vega: Should we use `tags/v0.71.4+fix` to make it clear which it is? I'm not sure about the `+` though.

davidsiska-vega (changes requested):

.env
TomMcL: Seems to work locally, let's give it a go

davidsiska-vega (approved):

PR #417, opened 2023-05-18
feat: at touch rl actions

TomMcL:
### Description

Adding further RL actions and fixing a scenario

### Testing

Manually tested

### Breaking Changes

None


parameter_results/MarkPriceUpdateFrequency/MarkPriceUpdateFrequency.ipynb
cdummett, line 1: I think there is data to be cleared in this notebook.

vega_sim/api/trading.py
```diff
@@ -240,7 +240,7 @@ def cancel_order(
     wallet_name: str,
     wallet: Wallet,
     market_id: str,
-    order_id: str,
+    order_id: Optional[str],
```
cdummett, line 243: Not sure I understand the purpose of making this optional?

vega_sim/scenario/configurable_market/scenario.py
```diff
@@ -98,6 +101,7 @@ def configure_agents(
         tag: str,
         market_config: Optional[MarketConfig] = None,
```
cdummett, line 102: I think when I made this I should have made this a `kwarg` instead, might be good to change here?

cdummett (commented):
Small comments - passing all tests though 👍

parameter_results/MarkPriceUpdateFrequency/MarkPriceUpdateFrequency.ipynb
cdummett: Beat me too it 👍

vega_sim/api/trading.py
TomMcL: If you pass as None it cancels everything on the market

vega_sim/scenario/configurable_market/scenario.py
TomMcL: If it's in the kwargs which are passed it will be assigned to here, otherwise default is used as normal, so expanding it specifically is fine

cdummett (approved):
LGTM

PR #418, opened 2023-05-19
feat: resource monitoring

cdummett:
### Description
Adds CPU Utilisation and Virtual Memory usage to the `Snitch`. Supporting plots added as well.

example

CPU monitoring supported on Darwin, Windows & Linux. Memory monitoring only supported on Darwin & Linux.

### Testing
Passing all tests locally.

### Breaking Changes
None

### Closes
None

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import psutil
+import platform
+
 import datetime
 import logging
 from dataclasses import dataclass
@@ -73,6 +76,17 @@ class MarketRegime:
     order_distribution_sell_kappa: float
     from_timepoint: int  # Inclusive
     thru_timepoint: int  # Inclusive
+
+
+@dataclass
+class ResourceData:
+    at_time: str
+    vega_cpu_per: float
+    vega_mem_rss: float
+    vega_mem_vms: float
+    datanode_cpu_per: float
+    datanode_mem_rss: float
+    datanode_mem_vms: float
 
 
 class TradeSignal(Enum):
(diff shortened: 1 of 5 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -2756,11 +2774,66 @@ def step(self, vega_state: VegaState):
...
+                    else 0,
+                )
+            )
         if self.additional_state_fn is not None:
             self.additional_states.append(
                 self.additional_state_fn(self.vega, self.agents)
             )
 
+    def _find_processes(self):
+        self.process_map.keys()
```
TomMcL: ?

vega_sim/scenario/common/agents.py
```diff
@@ -2756,11 +2774,66 @@ def step(self, vega_state: VegaState):
...
+        for p in psutil.process_iter():
+            if (p.name() in self.process_map.keys()) and (
+                p.status() in ["running", "sleeping"]
+            ):
```
TomMcL: I wonder whether there's something more robust we can do for this. This is currently going to break if someone ever runs two processes/there are leftover processes. But maybe that just has to be a caveat emptor for now.

TomMcL: Also I think we'd be better having the process map start empty then have the keys specified here directly, slightly clearer what it's for and avoids a case of `None`s in the dictionary being left over

cdummett: We could use the `cmdline` property to filter out any processes not started by a `VegaServiceNull` object?

Any other left-over "vega" and "data-node" processes started by other `VegaServiceNull` objects **_should_** be either `stopped` / `zombie` / `killed` by this point.

But yeah if we ever go for a multi-node setup in market-sim we would need to expand this to monitor all nodes.

TomMcL (approved):
lgtm

PR #424, opened 2023-05-26
feat: at touch mm

### LINKED ISSUE #423: Add at-the-touch-LP-agent
1. Add an agent that posts limit orders "at-the-touch". 
1. Add it to the fuzzing runs in some shape or form.

davidsiska-vega:
### Description

Will close #423 when added to the fuzzing runs.

### DIFF
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -3216,7 +3216,7 @@ class AtTheTouchMarketMaker(StateAgentWithWallet):
         self.current_position = 0
         self.order_size = 10 ** (-position_decimal_places)
         self.peg_offset = peg_offset
-        self.max_position = max_position 
+        self.max_position = max_position
 
         self.buy_order_reference = None
         self.sell_order_reference = None
@@ -3290,7 +3290,10 @@ class AtTheTouchMarketMaker(StateAgentWithWallet):
             self.sell_order_reference = None
 
         # If there is position within limits we place both buy and sell
-        if -self.max_position <= self.current_position and  self.current_position <= self.max_position:
+        if (
+            -self.max_position <= self.current_position
+            and self.current_position <= self.max_position
+        ):
             if self.buy_order_reference is None:
                 self.buy_order_reference = self.vega.submit_order(
                     trading_wallet=self.wallet_name,
--- a/vega_sim/scenario/common/agents.py
+++ b/vega_sim/scenario/common/agents.py
@@ -3199,7 +3199,7 @@ class AtTheTouchMarketMaker(StateAgentWithWallet):
         position_decimal_places: int = 0,
         tag: str = "",
         wallet_name: str = None,
-        peg_offset = 1,
+        peg_offset=1,
     ):
         super().__init__(wallet_name=wallet_name, key_name=key_name, tag=tag)
         self.initial_asset_mint = initial_asset_mint
@@ -3213,11 +3213,11 @@ class AtTheTouchMarketMaker(StateAgentWithWallet):
         self.asset_name = asset_name
 
         self.current_position = 0
-        self.order_size = 10**(-position_decimal_places)
+        self.order_size = 10 ** (-position_decimal_places)
         self.peg_offset = peg_offset
 
-        self.buy_order_reference = None 
-        self.sell_order_reference = None 
+        self.buy_order_reference = None
+        self.sell_order_reference = None
 
     def initialise(
         self,
@@ -3245,11 +3245,8 @@ class AtTheTouchMarketMaker(StateAgentWithWallet):
 
         self._update_state(current_step=self.current_step)
 
-        
     def step(self, vega_state: VegaState):
         self.current_step += 1
-        
-        # self._update_state(current_step=self.current_step)
 
         # Each step, get position
         position = self.vega.positions_by_market(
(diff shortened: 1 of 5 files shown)


vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+                sell_seen = True
+        if not buy_seen:
+            self.buy_order_reference = None
+        if not sell_seen:
+            self.sell_order_reference = None
+
+        # If there is position within limits we place both buy and sell
+        if (
+            -self.max_position <= self.current_position
+            and self.current_position <= self.max_position
```
TomMcL: `abs(self.current_position) <= self.max_position`

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+        tag: str = "",
+        wallet_name: str = None,
+        peg_offset=1,
+        max_position=1,
+    ):
+        super().__init__(wallet_name=wallet_name, key_name=key_name, tag=tag)
+        self.initial_asset_mint = initial_asset_mint
+        self.mdp = market_decimal_places
+        self.adp = asset_decimal_places
+        self.pdp = position_decimal_places
```
TomMcL: I don't think we need these?

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+
+        self.current_step = 0
+
+        self.market_name = market_name
+        self.asset_name = asset_name
+
+        self.current_position = 0
+        self.order_size = 10 ** (-position_decimal_places)
```
TomMcL: Why the tiny order size?

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+                    trading_key=self.key_name,
+                    market_id=self.market_id,
+                    order_type="TYPE_LIMIT",
+                    time_in_force="TIME_IN_FORCE_GTC",
+                    side="SIDE_BUY",
+                    volume=self.order_size,
+                    price=None,
+                    expires_at=None,
+                    pegged_order=PeggedOrder(
+                        reference="PEGGED_REFERENCE_MID", offset=self.peg_offset
```
TomMcL: These should be pegged to best bid/ask no?

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+                        side="SIDE_BUY",
+                        volume=self.order_size,
+                        price=None,
+                        expires_at=None,
+                        pegged_order=PeggedOrder(
+                            reference="PEGGED_REFERENCE_MID", offset=self.peg_offset
+                        ),
+                        order_ref=str(uuid.uuid4()),
+                        wait=True,
+                    )
```
TomMcL: This logic feels duplicative, thinking how to simplify

TomMcL (changes requested):

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,209 @@ def initialise(
...
+                    self.vega.wallet.public_key(
+                        wallet_name=self.wallet_name, name=self.key_name
+                    ),
+                    {},
+                )
+                .values()
+            )
+            if self.market_id in vega_state.market_state
+            else []
+        )
```
TomMcL, line 3279: `orders_for_party_from_feed`?

vega_sim/scenario/common/agents.py
TomMcL: I think this could be simplified down to something like:

```
place_buy = self.current_position <= self.max_position
place_sell = self.current_position >= -self.max_position

if self.buy_order_reference is None and place_buy:
   place_order
elif self.buy_order_reference is not None and not place_buy:
   cancel_order

vice versa for sell
```

vega_sim/scenario/common/agents.py
cdummett: I thought along the lines of this [thread](https://vegaprotocol.slack.com/archives/C03P2B6A3EH/p1678889198337989?thread_ts=1678887466.901759&cid=C03P2B6A3EH) we use `vega_state` where possible?

For checking an agents own orders though I can definitely see an argument for them being allowed to get the latest order states by using the `from_feed` method.

vega_sim/scenario/common/agents.py
cdummett: Think it was originally a quick work around to not have to consider the partially filled order case - with the changes to how we determine whether an order is live and the max position logic I don't think we need to worry about partially filled orders any more.

@davidsiska-vega as long as you are happy with us not "refreshing" the volume of the order on a partial fill and only doing so once the order is fully filled (of course still following the max position conditions).

davidsiska-vega: That may be fine ;-).

vega_sim/scenario/common/agents.py
cdummett: Not sure I could see an argument for both being useful, all okay if I make it configurable @TomMcL ?

TomMcL: Configurable works too, at the touch generally means best bid/ask though so that should be default :)

vega_sim/scenario/common/agents.py
TomMcL: That is true, let's stick to being good citizens with `vega_state`

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,158 @@ def initialise(
...
+        if not sell_seen:
+            self.sell_order_reference = None
+
+        place_buy = self.current_position < self.max_position
+        place_sell = self.current_position > -self.max_position
+
+        if self.buy_order_reference is None and place_buy:
+            self.buy_order_reference = self._place_order(side="SIDE_BUY")
+        elif self.buy_order_reference is not None and not place_buy:
+            self._cancel_order(order_id="SIDE_BUY")
```
TomMcL: Need to fix this kwarg ;)

vega_sim/scenario/common/agents.py
```diff
@@ -3175,3 +3177,158 @@ def initialise(
...
+            price=None,
+            expires_at=None,
+            pegged_order=PeggedOrder(
+                reference=self.buy_peg_reference
+                if side == "SIDE_BUY"
+                else self.sell_peg_reference,
+                offset=self.peg_offset,
+            ),
+            order_ref=str(uuid.uuid4()),
+            wait=True,
```
TomMcL: set this to `False`

vega_sim/scenario/configurable_market/scenario.py
```diff
@@ -152,6 +153,20 @@ def configure_agents(
...
+        at_the_touch_mm = AtTheTouchMarketMaker(
+            key_name="AT_THE_TOUCH_MM",
+            wallet_name="MM",
+            initial_asset_mint=1e9,
+            market_name=self.market_name,
+            asset_name=self.asset_name,
+            market_decimal_places=market_config.decimal_places,
+            position_decimal_places=market_config.position_decimal_places,
+            asset_decimal_places=self.asset_decimal,
+            peg_offset=30 * 10 ** (-market_config.decimal_places),
```
TomMcL: If we can set `peg_offset` in both this and below to `0` I think we're good to go

TomMcL (approved):

PR #426, opened 2023-05-26
feat: add governance transfer method

cdummett:
### Description

PR makes necessary changes to implement governance transfers into market-sim.

- adds a simple `propose_transfer` method
- adds a `test_governance_transfer` test for simple method

To support fuzzing of complicated commands PR introduces the `builders` package and `fuzzers` module.

#### `builders`
- package contains functions for building protobuf messages
- package structured to match vega proto structure so docs align
- functions handle conversion of vms types to vega types
- functions have NO validation

#### `fuzzers`
- package contains functions for building fuzzed protobuf messages
- each field has a random probability of being "valid"
- functions fuzz in a top down approach, i.e. the deepest messages are fuzzed first and then used to determine the validity of shallower fields.

TODO: 
- A future PR will align service methods and API functions to use builder functions.
- A future PR will audit the entirety of market-sim to make sure anything above the service layer uses market-sim types.

### Testing
All tests passing locally


vega_sim/builders/governance.py
```diff
@@ -0,0 +1,161 @@
...
+
+Attributes:
+    logger (logging.Logger): module level logger
+"""
+import logging
+import datetime
+
+import vega_sim.proto.vega as vega_protos
+
+from typing import *
```
TomMcL: Best not to have `import *` if we can avoid it.

vega_sim/scenario/fuzzed_markets/scenario.py
```diff
@@ -332,11 +333,11 @@ def configure_agents(
                     referrer_wallet_name="REFERRERS",
                 )
                 for i_agent, i_referrer in itertools.product(
-                    range(5), range(num_referrers)
+                    range(1), range(num_referrers)
```
TomMcL: These reductions meant to still be here?

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -1308,170 +1276,130 @@ def step(self, vega_state):
...
+                    approve_proposal=True,
+                    wallet_name=self.wallet_name,
+                )
+                continue
+            except (HTTPError, ProposalNotAcceptedError):
+                self.accepted_proposals += 1
+                continue
+
+    def finalise(self):
+        logging.info(
```
TomMcL: Maybe make this debug

TomMcL (changes requested):
Couple of small things

TomMcL (approved):

PR #435, opened 2023-06-12
feat: add gtt liquidation example

cdummett:
#### Summary:

PR adds a visualisation example for how expiring `GTT` orders not triggering margin recalculations may result in an unfair liquidation. Related to bug [#8482](https://github.com/vegaprotocol/vega/issues/8482)

#### Method:

1. `PARTY A` takes on a short position in `MARKET A` - this is the position to be liquidated
2. `PARTY A` takes on a short position in `MARKET B`
3. `PARTY A` places a large `GTT` sell order in `MARKET B` which we immediately expire

At this point `PARTY A` will have a large proportion of their funds locked in the margin account for `MARKET B `as a result of the expired `GTT` short order. No funds were released from this account when the order expired as margin levels were not recalculated and `PARTY A` had an open position in `MARKET B`. As long as margin calculations are not triggered for `PARTY A` in `MARKET B`, the funds will not be released.

4. Gradually increase the mark price in `MARKET B`

Eventually `PARTY A` will not have enough funds left in their general account to support their position in `MARKET A`. The funds currently locked in the margin account of `MARKET B` will also not be used to support their position in `MARKET A` despite the fact `PARTY A` has no open orders and a small position in `MARKET B`.

### Example
Run the following command:
```
python -m examples.visualisations.gtt_liquidation --console
```
The logs will show `PARTY A` is liquidated in `MARKET_A` with marginal price movements and that the party still has a large quantity of remaining funds. These funds are locked in the margin account of `MARKET B` despite the party having  no open orders and a small position in that market.

### Testing
Passing all tests.

### Breaking Changes
None

### Closes
None

### DIFF
--- a/examples/visualisations/gtt_liquidation.py
+++ b/examples/visualisations/gtt_liquidation.py
@@ -1,0 +1,181 @@
+"""gtt_liquidation.py
+
+Script visualises an example where a large expired GTT order causes a party to
+be closed out as margin levels are not recalculated after a GTT order expires.
+
+Method:
+
+1. PARTY A takes on a short position in MARKET A - this is the position to be liquidated
+2. PARTY A takes on a short position in MARKET B
+3. PARTY A places a large GTT sell order in MARKET B which we immediately expire
+
+At this point PARTY A wil have a large proportion of their funds locked in the margin
+account for MARKET B as a result of the expired GTT short order. No funds were released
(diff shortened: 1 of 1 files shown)


cdummett:
Closing for same reasons mentioned in https://github.com/vegaprotocol/vega/issues/8482

PR #437, opened 2023-06-20
chore: Updating to latest release tag

TomMcL:
### Description

Updating market sim and ports to latest release tag for core

### Testing

Integration tests

### Breaking Changes

None


cdummett (changes requested):
Need to update the tag in the `Jenkinsfile` as well?

TomMcL:
@cdummett the dangers of a manual update, good catch

cdummett (approved):

PR #438, opened 2023-06-21
Allow specifying a parital port_config

mattrussell36:
### Description

Adjusts port config setup so that you can set a single port and not have to provide _all_ the ports

### Testing
Added an integration test `test_port_config.py`

### Breaking Changes
I don't think so, but you might be able to remove some ports if you use port_config and only need to set a few of them

### Closes
N/A

### DIFF
--- a/vega_sim/null_service.py
+++ b/vega_sim/null_service.py
@@ -709,7 +709,7 @@ class VegaServiceNull(VegaService):
                 curr_ports = set(
                     [getattr(self, port) for port in self.PORT_TO_FIELD_MAP.values()]
                 )
-                setattr(self,name, find_free_port(curr_ports))
+                setattr(self, name, find_free_port(curr_ports))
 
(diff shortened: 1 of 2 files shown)


vega_sim/null_service.py
```diff
@@ -709,6 +689,28 @@ def _generate_port_config(self) -> Dict[Ports, int]:
             Ports.CONSOLE: self.console_port,
         }
 
+    # set ports from port_config or alternatively find a free port
+    # to use
+    def _assign_ports(self, port_config):
```
TomMcL:
```suggestion
    def _assign_ports(self, port_config: Optional[Dict[Ports, int]]):
```

TomMcL (changes requested):

vega_sim/null_service.py
mattrussell36: Done

TomMcL (approved):

PR #439, opened 2023-06-26
feat: Updating versions to be less prescriptive and allow python 3.11

TomMcL:
### Description

Updating python to allow version 3.11, also updating to a different package for openapi code to allow later versions of protoc. Also generically making most packages more flexible on versioning

### Testing

Integration tests should cover

### Breaking Changes

Should be none


cdummett (approved):
LGTM - all tests passing on both 3.10 and 3.11

PR #456, opened 2023-07-18
feat: Removing waiting from at-touch market maker

TomMcL:
### Description

Simple change to remove at-touch market maker's wait from order placing which occasionally fails if they run out of money

### Testing

Integration tests will cover

### Breaking Changes

None


cdummett (approved):
LGTM 👍

PR #460, opened 2023-07-19
feat: Adding ability to have single reward shared across agents in rl

TomMcL:
### Description

Adding ability in the RL framework to pass a single reward function which is shared across all agents

### Testing

Ran locally, worked

### Breaking Changes

Should be none


Jiajia-Cui (approved):
LGTM

PR #461, opened 2023-07-20
feat: iceberg order

cdummett:
### Description
PR updates market-sim order submissions with iceberg functionality.

### Testing
Test added and passing locally.

### Breaking Changes
None


vega_sim/service.py
```diff
@@ -794,6 +800,12 @@ def submit_order(
             else:
                 logger.debug(msg)
                 return
+        if (peak_size is None and minimum_visible_size is not None) or (
+            peak_size is not None and minimum_visible_size is None
+        ):
+            raise ValueError(
+                "Both 'peak_size' and 'minimum_visible_size' must be specified or none at all."
+            )
```
TomMcL, line 808: Side note and not at all actually a change request, I've been wondering to myself recently whether this is a better pattern or whether it's better to create a custom DataType which requires both peak_size and minimum_visible_size to be set on it.

TomMcL (approved):
lgtm

vega_sim/service.py
cdummett: Oh that is a nicer idea, I always to-and-fro on what the best way to do this as well. 

This way is easy enough when you only have two fields but when you have lots it becomes messy so updating all of these to a dataclass would be nice.

Also probably one step too far... but I keep thinking having a 1-1 between market-sim data classes and vega protos  might be useful. Each class would then contain the methods for switching between the two (converting padded ints to floats etc.) and we could add the above validation to the class as well?

Mostly thinking about it because the `data.py` methods return a mix of `dataclasses`, `namedtuples` and `protos` at the moment which gives me a headache 😆

Issue #442, opened 2023-06-28
Add successor markets to a fuzzing run


davidsiska-vega:
We would like to create a new "scenario" (term used loosely) that exercises the successor market feature. 

Something along the lines:
[ ] a sequence of (at least 3) successor markets gets proposed and enacted. 
[ ] at least three rational LP agents commit liquidity to the 1st market, 1st market settles, for the 2nd market one of them keeps same commitment, one reduces, one increases... 
[ ] price taking agents (informed, random, degen) trade on the market 
[ ] we have some visualisation for the virtual stakes across the market sequence
[ ] this "scenario" is included in the nightly fuzzing run

PR #463, opened 2023-07-24
feat: successor fuzz

TomMcL:
### Description

Adding successor market functionality, adding it to fuzzing run and updating plotting to handle it

### Testing

Ran locally, included in fuzzing runs

### Breaking Changes

Should be none

### Closes

#442


vega_sim/api/market.py
```diff
@@ -124,6 +126,8 @@ def load(self, opt: Optional[str] = None):
         self.quadratic_slippage_factor = str(config["quadratic_slippage_factor"])
         self.metadata = config["metadata"]
 
+        self.successor = Successor(opt=config["successor"])
```
cdummett: I think since `successor` is an optional field in the `NewMarketConfiguration` we should be able to handle `None` values here:
```suggestion
        self.successor = Successor(opt=config["successor"]) if config["successor"] is not None else None
```

vega_sim/api/market.py
```diff
@@ -111,6 +112,7 @@ class MarketConfig(Config):
             "lp_price_range": 0.5,
             "linear_slippage_factor": 1e-3,
             "quadratic_slippage_factor": 0,
+            "successor": "default",
```
cdummett: And then maybe default to `None` here?
```suggestion
            "successor": None,
```

cdummett (commented):
Everything is passing locally for me, not sure what that failure is in Jenkins.

Looks good, just one point about handling optional args in the market config. If you want to keep as is I'm very fine with that and will update how I'm doing it for the Spot stuff to match.

cdummett (approved):
lgtm

PR #468, opened 2023-07-27
feat: Adding better float handling to conversion

TomMcL:
### Description

Better decimal handling to avoid weird rounding when converting to padded ints

### Testing

Added unit tests, breaking before change and working after

### Breaking Changes

Should be none


davidsiska-vega (approved):
Sneaky bug. How did you find it?

TomMcL:
> Sneaky bug. How did you find it?

Frontend tests caring about the order price with very specific numbers :D

PR #469, opened 2023-07-28
feat: update spam policies for null service

daniel1302:
# Description TBD

### Closes

close vegaprotocol/devops-infra#1531


daniel1302 (commented):
won't merge because there is a newer pr that fixes the same issue and even more.

daniel1302:
WONT m

PR #471, opened 2023-07-29
feat: set of fixes and improvements for vega-market-sim

daniel1302:
# Description

### Description

I really need this pipeline to test my nullchain changes in the core.

#### Changes

- Added a generic function to the `tools.retry` function that retries the given code and returns the result or throws an exception when all of the attempts fail.
- Added support for the following parameters: `PARALLEL_WORKERS`, `TEST_FUNCTION`, and `LOG_LEVEL` - They allow driving the integration tests script. Jenkins now has support for it - see the related PR.
- Added support for the parallel run of the scenarios
    - There is a `pytest-dist` library that has been already included in the `poetry.lock`. I have tested those PRs against that run as well
    - There is a limitation when we use parallel run of the scenarios, where pytest does not print logs to the STDOUT anymore. I have added `conftest.py`, and some params in the `pytest.ini`. Those settings enable writing test logs into the file. Now files are saved in the `Jenkins artifacts (you can confirm it here: https://jenkins.ops.vega.xyz/job/common/job/vega-market-sim/2607/artifact/test_logs/ ) - one file per worker.
- Disabled the `test_rl_run` test for the integration run. It takes a very long time to run this single scenario. There is a following issue here(https://github.com/vegaprotocol/vega-market-sim/issues/472) to run it
- The biggest problem was a huge memory leak that caused us to need 64GB of RAM on the server to fit the run. Now we need 16 GB max, with 10 parallel runs the pipeline uses 12 GB. On docker, it could work(I am not sure if it was) because docker manages fork/workers differently. When We disabled docker it started causing huge issues. I have implemented a fix, but this fix is not ideal. I have left a comment block that describes limitations. Here is a follow-up issue to fix it properly: TBD...
- Changed sleep time, there is no reason to sleep `0.0001s` and send up to 1000 requests per second. New sleep is 0.05 and time increases exponentially. Moreover, I have also decreased the number of tries - see details below. 
- Increased time for waiting for the proposal from 0.8 sec to about 2.8 sec. The timeout formula is now `0.05 * 1.1**i`

#### Sleep Changes

- The old sleep formula was: `[0.0005 * 1.01**attempt for attempt in range(625)]`
- The new formula is:  `[0.05 * 1.03**attempt for attempt in range(100)]`

Max wait:
- for new formula: `sum([0.05 * 1.03**i for i in range(100)]) == 30.364386634760475` - 30.4 sec.
- for old formula: `sum([0.0005 * 1.01**i for i in range(625)]) == 25.05895385147325` - 25.1 sec

The difference is that instead of `625` we send max `100` requests.
- When we were sleeping `0.0005` seconds it was sending about **250 requests**, (`sum([0.0005 * 1.01**i for i in range(260)]) - 0.6145492739027707`) sometimes more - this is my observation
- Now it is sending usually 1 request, in some scenarios 3-to 4 requests.

Because we were sending thousands of requests without reason we saw a lot of RPC Connection issues.

### Testing

- https://jenkins.ops.vega.xyz/job/common/job/vega-market-sim/2605/
- https://jenkins.ops.vega.xyz/job/common/job/vega-market-sim/2606/

image

### Breaking Changes

No breaking changes.

### Closes

close vegaprotocol/devops-infra#1531 .

# Related PRs:

- https://github.com/vegaprotocol/jenkins-shared-library/pull/571
- https://github.com/vegaprotocol/vega-market-sim/pull/471

### DIFF
--- a/tests/integration/utils/fixtures.py
+++ b/tests/integration/utils/fixtures.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+import logging
 from typing import Optional
 
 import pytest
@@ -137,7 +138,7 @@ def build_basic_market(
     vega.wait_for_total_catchup()
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def vega_service():
     with VegaServiceNull(
         warn_on_raw_data_access=False,
@@ -147,15 +148,17 @@ def vega_service():
         listen_for_high_volume_stream_updates=False,
     ) as vega:
         yield vega
+    logging.debug("vega_service theardown")
+    
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def vega_service_with_market(vega_service):
     build_basic_market(vega_service, initial_price=0.3)
     return vega_service
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def vega_service_with_high_volume():
     with VegaServiceNull(
         warn_on_raw_data_access=False,
@@ -165,9 +168,10 @@ def vega_service_with_high_volume():
         listen_for_high_volume_stream_updates=True,
     ) as vega:
         yield vega
+    logging.debug("vega_service_with_high_volume theardown")
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
 def vega_service_with_high_volume_with_market(vega_service_with_high_volume):
     build_basic_market(vega_service_with_high_volume, initial_price=0.3)
     return vega_service_with_high_volume
--- a/tests/integration/utils/fixtures.py
+++ b/tests/integration/utils/fixtures.py
@@ -148,7 +148,7 @@ def vega_service():
         listen_for_high_volume_stream_updates=False,
     ) as vega:
         yield vega
-    logging.debug("vega_service theardown")
+    logging.debug("vega_service teardown")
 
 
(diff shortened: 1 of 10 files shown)


vega_sim/api/helpers.py
```diff
@@ -56,7 +56,7 @@ def num_from_padded_int(to_convert: Union[str, int], decimals: int) -> float:
 def wait_for_datanode_sync(
     trading_data_client: VegaTradingDataClientV2,
     core_data_client: VegaCoreClient,
-    max_retries: int = 650,
+    max_retries: int = 100,
```
daniel1302, line 59: fix comments below

tests/integration/utils/fixtures.py
```diff
@@ -147,15 +148,16 @@ def vega_service():
         listen_for_high_volume_stream_updates=False,
     ) as vega:
         yield vega
+    logging.debug("vega_service theardown")
 
 
-@pytest.fixture
+@pytest.fixture(scope="function")
```
TomMcL, line 154: function should be the default scope for these, but no harm being explicit!

TomMcL (approved):
LGTM, need to reverse engineer how this is fixing datanode still, but I'll take a working node!

PR #492, opened 2023-08-23
feat: SLA Updates

TomMcL:
### Description

Updates to allow SLA changes to work within market sim

### Testing

Running integration tests

### Breaking Changes

Various changes to market setup and LP creation. Mainly LP bots would be affected but these should be changed too.


cdummett (dismissed):
LGTM

davidsiska-vega (approved):

vega_sim/scenario/common/agents.py
```diff
@@ -1627,7 +1617,15 @@ def _calculate_price_volume_levels(
...
         level_price[level_price < 1 / 10**self.mdp] = 1 / 10**self.mdp
 
-        return [MMOrder(vol, price) for vol, price in zip(level_vol, level_price)]
+        orders = [MMOrder(vol, price) for vol, price in zip(level_vol, level_price)]
+        total_liq_supp = sum([order.size * order.price for order in orders])
+        remaining = self.commitment_amount - total_liq_supp
+
+        if remaining > 0:
+            next_price = orders[-1].price + mult_factor * self.tick_spacing
+            orders.append(MMOrder(remaining / next_price, next_price))
```
cdummett, line 1626: I think since we scale the orders to meet the LP commitment after we create the shape this might be in the wrong place?

Would it not be better in the ShapedMarketMaker `_scale_orders` method?

PR #498, opened 2023-08-30
feat: pipe logging to single process

TomMcL:
### Description

Piping all log outputs from process management process to the main process to allow output to files etc

### Testing

Ran locally, output saved correctly

### Breaking Changes

Should for the most part act as before. Saved log files may have a couple more lines in


tests/integration/utils/fixtures.py
```diff
@@ -144,7 +144,7 @@ def vega_service():
         retain_log_files=True,
         transactions_per_block=1,
         listen_for_high_volume_stream_updates=False,
-        use_full_vega_wallet=True,
+        use_full_vega_wallet=False,
```
cdummett, line 147: Is this intentional? Think we switched from `False` -> `True` yesterday and wonder if this is a conflict.

tests/integration/utils/fixtures.py
```diff
@@ -144,7 +144,7 @@ def vega_service():
         retain_log_files=True,
         transactions_per_block=1,
         listen_for_high_volume_stream_updates=False,
-        use_full_vega_wallet=True,
```
cdummett, line 147: Is this intentional? Think we switched from `False` -> `True` yesterday and wonder if this is a conflict.

tests/integration/utils/fixtures.py
```diff
@@ -164,7 +164,7 @@ def vega_service_with_high_volume():
         retain_log_files=True,
         transactions_per_block=1,
         listen_for_high_volume_stream_updates=True,
-        use_full_vega_wallet=True,
+        use_full_vega_wallet=False,
```
cdummett, line 167: ditto

tests/integration/utils/fixtures.py
TomMcL: No the switch to `True` was the non-deliberate part 😁

cdummett (approved):

PR #504, opened 2023-09-05
feat: handle genesis changes

TomMcL:
### Description

Fixes for core changes and better shutdown to avoid infinite waits

### Testing

Tested locally. Now works and correctly closes on errors

### Breaking Changes

None


cdummett (commented):
Looks good but could we add the new network parameters to the genesis file as a part of this PR with the proposed mainnet values?

cdummett (approved):

PR #505, opened 2023-09-05
feat: SLA scenario and analysis plots

### LINKED ISSUE #493: SLA agent support and visualisations
Some validation / visualisation runs for SLA (will need more agents): 

1. An LP A posts volume that just meets the SLA, no more, no less(*) vs. an LP B that is halfway from the SLA to full volume on the book all the time (so if the SLA on the market is 90% we have SLA posting volume 90% of the time and SLA B posting volume 95% of the time). They post equally competitive volume. Visualise their LP fee income for each epoch.
2. An LP A posts volume that just meets the SLA, no more, no less. An LP B posts volume that is SLA - 10% of the time, i.e. if. the market SLA is 95% LP B posts only 85% of the time. Visualise their LP fee income for each epoch. Visualise their LP bond balances and virtual stakes. We should see LP B’s bond bleeding to 0 and no LP fee income for them.
3. LPs A and B post volume that meets the SLA exactly but LP A’s volume is priced more competitively. Visualise their LP fee income for each epoch.
4. LP A posts volume that meets the SLA. The feeCalculationTimeStep (when it is a netparam(**)) is set to 0, vs 1min vs 1h vs 6h. Visualise the LP fee income for each epoch for each value and check it looks sane. 

(*) One option to code this is that every period of time the agent flips a biased coin and decides whether to have volume on the book or not.
(**) I think at the moment it’s a market param but that should change for validation purposes.

cdummett:
### Description
PR adds an `SLAScenario` which allows the configuration of SLA parameters as well as the configuration of a number of competing liquidity providers (plotting tools are optimised for 2 LPs so this is the default).

Three pre-configured scenarios have been added to the registry:

- `sla_a` - LP 1 fully meets their commitment, LP2 only just meets their commitment. Equally competitive orders.
- `sla_b` - LP 1 only just meets their commitment, LP2 does not meet their commitment. Equally competitive orders.
- `sla_c` - LP 1 only just meets their commitment, LP2 only just meets their commitment. LP A orders more competitive.

To support the analysis, plotting tools added.

#### Scenario `sla_a` example plots: 
LP1 receives all their fees and the majority of LP2's fees who only just met their commitment.
sla_a

#### Scenario `sla_b` example plots: 
LP1 receives no fees as a net distribute as they only just met their commitment but instead receives all the fees as a bonus as they were the only LP who met their commitment.
sla_b

#### Scenario `sla_c` example plots: 
Both LPs only just meet their commitment, LP2 accumulates and receives less fees as their orders were less competitively priced.
sla_c

### Testing
Passing all tests locally.

To test yourself.

1. `vega_sim.scenario.adhoc -s sla_a --output`
2. `vega_sim.tools.scenario_plots --sla --show`

### Breaking Changes
None.

Things to consider though:
- `Snitch` now supports a `final_extraction_fn` which can be run in its finalise step.
- `ShapedMarketMaker` (and child classes) now have functionality to specify a `supplied_amount` independent of the `commitment_amount`. Allows the creation of MMs which are not liquidity providers.

### Closes
Closes #493


davidsiska-vega (dismissed):
I just eyeballed the plots but it's looking good!

vega_sim/api/data.py
```diff
@@ -41,26 +41,35 @@ class MissingMarketError(Exception):
...
+class Order:
+    price: float
+    size: float
+    id: str
+    reference: str
+    side: vega_protos.vega.Side
+    status: vega_protos.vega.Order.Status
+    remaining: float
+    time_in_force: vega_protos.vega.Order.TimeInForce
+    order_type: vega_protos.vega.Order.TimeInForce
```
TomMcL: copy and paste going rogue :)

vega_sim/api/data.py
```diff
@@ -343,6 +352,23 @@ def _margin_level_from_proto(
     )
 
 
+def _iceberg_order_from_proto(
+    iceberg_order: vega_protos.vega.IcebergOrder,
+    decimal_spec: DecimalSpec,
+) -> Order:
```
TomMcL:
```suggestion
) -> IcebergOrder:
```

vega_sim/api/data.py
```diff
@@ -363,6 +389,9 @@ def _order_from_proto(
         updated_at=order.updated_at,
         version=order.version,
         market_id=order.market_id,
+        iceberg_order=_iceberg_order_from_proto(order.iceberg_order, decimal_spec)
+        if order.HasField("iceberg_order")
+        else None,
     )
```
TomMcL, line 395: At some point these could be made into `classmethod`s on the data classes themselves, though probably not now

vega_sim/service.py
```diff
@@ -2048,15 +2050,27 @@ def create_order_submission(
...
     def submit_instructions(
         self,
         key_name: str,
         wallet_name: Optional[str] = None,
-        cancellations: Optional[List[OrderCancellation]] = None,
-        amendments: Optional[List[OrderAmendment]] = None,
-        submissions: Optional[List[OrderSubmission]] = None,
+        cancellations: Optional[List[OrderCancellation]] = [],
+        amendments: Optional[List[OrderAmendment]] = [],
+        submissions: Optional[List[OrderSubmission]] = [],
```
TomMcL: These should still be `None`. Set them inside the function with `cancellations = cancellations if cancellations is not None else []`. (as they are here these lists are mutable, and use the same object for every call of the function :) )

vega_sim/api/trading.py
```diff
@@ -547,7 +547,6 @@ def order_submission(
         type=order_type,
         reduce_only=reduce_only,
         post_only=post_only,
-        iceberg_opts=iceberg_opts,
```
TomMcL, line 550: Why is this being deleted?

TomMcL (changes requested):

vega_sim/service.py
cdummett: Ah yes thank you! Always forget this one.

vega_sim/api/trading.py
cdummett: We were already setting the opts a few lines below with this and didn't seem worth the duplication.
```
    if iceberg_opts is not None:
        command.iceberg_opts.CopyFrom(iceberg_opts)
```

TomMcL (dismissed):

TomMcL (approved):

PR #515, opened 2023-09-26
feat: perf improvements

TomMcL:
### Description

Various performance improvements, mainly aimed at speeding up the long fuzzing run where data starts to build up and queries slow

### Testing

Ran fuzzing locally and all looked fine

### Breaking Changes

Should be none


vega_sim/service.py
```diff
@@ -1169,17 +1170,28 @@ def settle_market(
...
     ) -> data.PartyMarketAccount:
         """Output money in general accounts/margin accounts/bond accounts (if exists)
         of a party."""
-        return data.party_account(
-            self.wallet.public_key(wallet_name=wallet_name, name=key_name),
-            asset_id=asset_id,
+        if market_id is None and asset_id is None:
+            raise Exception("Either market or asset must be passed")
+        if market_id is not None and asset_id is not None:
+            raise Exception("Only one of market or asset can be passed")
```
cdummett, line 1182: Think these might be causing some of the failures, what's the purpose of this check?

TomMcL: Failures all fixed up now. There are a few different ways one could be wanting data from this and it's tricky to disambiguate them all! This was pretty much just removing one ambiguity (where you want everything, which I've added a function for in local data cache but haven't bothered piping through yet)

cdummett (dismissed):

cdummett (approved):

PR #517, opened 2023-09-29
feat: Adding port setting for metrics to fuzz test run

TomMcL:
### Description

Adding ability to manually configure port for metrics to a known fixed port on fuzz testing runs

### Testing

Ran locally and port was as configured

### Breaking Changes

run_fuzz_test script is changed so proc files attempting to call it will need slight updating


davidsiska-vega (approved):
Looks fine, from my limited understanding. 

Passing tests locally.

cdummett (approved):

PR #518, opened 2023-09-29
feat: integrate the new reward metrics into the 

cdummett:
### Description
PR integrates the new reward metrics into market-sim.

### Testing
Passing all tests locally.


tests/integration/utils/fixtures.py
```diff
@@ -167,7 +167,7 @@ def vega_service():
         retain_log_files=True,
         transactions_per_block=1,
         listen_for_high_volume_stream_updates=False,
-        use_full_vega_wallet=False,
```
TomMcL: Guessing this isn't deliberate?

vega_sim/service.py
```diff
@@ -2373,6 +2400,73 @@ def recurring_transfer(
             recurring=recurring_transfer,
         )
 
+    def dispatch_strategy(
```
TomMcL: This should be in `data.py` rather than a method on the class

tests/integration/utils/fixtures.py
cdummett: Yep, good catch thanks

TomMcL (approved):

PR #520, opened 2023-10-05
feat: integrate teams into market-sim

cdummett:
### Description
PR integrates APIs to cover referral sets and teams.

The fuzzing test now passes the relevant fields to create a team from the referral set.

### Testing
- Tests added for new APIs.
- Tests added to check the coverage of vega APIs - tests are marked as `api` tests which are not run as part of integration tests.

To check coverage

```
make test_api
```

### Breaking Changes
None

### Closes
None

### DIFF
--- a/vega_sim/api/data.py
+++ b/vega_sim/api/data.py
@@ -281,11 +281,11 @@ class ReferralSetReferee:
 @dataclass(frozen=True)
 class ReferralSetStats:
     at_epoch: int
-    referral_set_running_notional_taker_volume: int
+    referral_set_running_notional_taker_volume: float
     party_id: str
     discount_factor: float
     reward_factor: float
-    epoch_notional_taker_volume: int
+    epoch_notional_taker_volume: float
 
 
 @dataclass(frozen=True)
@@ -2222,7 +2222,7 @@ def list_teams(
     data_client: vac.trading_data_grpc_v2,
     team_id: Optional[str] = None,
     party_id: Optional[str] = None,
-) -> List[Team]:
+) -> Dict[str, Team]:
     response = data_raw.list_teams(
         data_client=data_client, team_id=team_id, party_id=party_id
     )
--- a/vega_sim/api/data.py
+++ b/vega_sim/api/data.py
@@ -360,6 +360,33 @@ class VolumeDiscountProgram:
     ended_at: int
 
 
+@dataclass(frozen=True)
+class Team:
+    team_id: str
+    referrer: str
+    name: str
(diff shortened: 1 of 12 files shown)


vega_sim/api/data.py
```diff
@@ -278,6 +278,39 @@ class ReferralSetReferee:
...
 
 
+@dataclass(frozen=True)
+class ReferralSetStats:
+    at_epoch: int
+    referral_set_running_notional_taker_volume: int
+    party_id: str
+    discount_factor: float
+    reward_factor: float
+    epoch_notional_taker_volume: int
```
TomMcL: If this is in `data` these should be converted to real numbers so should be floats I think?

vega_sim/api/data.py
```diff
@@ -2013,3 +2216,37 @@ def dispatch_strategy(
...
+
+def list_teams(
+    data_client: vac.trading_data_grpc_v2,
+    team_id: Optional[str] = None,
+    party_id: Optional[str] = None,
+) -> List[Team]:
+    response = data_raw.list_teams(
+        data_client=data_client, team_id=team_id, party_id=party_id
+    )
+    return {team.team_id: _team_from_proto(team=team) for team in response}
```
TomMcL, line 2229: This isn't a list :)

vega_sim/api/trading.py
```diff
@@ -691,16 +702,27 @@ def create_referral_set(
...
+    if any(arg is not None for arg in [name, team_url, avatar_url, closed]):
+        is_team = True
+    else:
+        is_team = False
     command = vega_protos.commands.v1.commands.UpdateReferralSet(
         is_team=is_team,
     )
     if is_team:
+        if not all(arg is not None for arg in [name, team_url, avatar_url, closed]):
+            raise ValueError("If one team arg passed, all team args must be passed.")
```
TomMcL, line 725: I'd be tempted to still default the URLs since we're unlikely to ever care, unless there's a reason not to?

TomMcL (changes requested):

TomMcL:
I'm not *entirely* sure on the value of the api tests given that I'm not sure covering every proto etc is an end goal, but they don't hurt so let's add them in!

vega_sim/api/data.py
cdummett: Yeah these will be quantum values so need to do any conversion but you are right these should be floats anyway, quantum values don't have to be floats 👍

TomMcL (approved):

PR #521, opened 2023-10-06
feat: integrate stop orders into market-sim

cdummett:
### Description
PR implements stop orders into market-sim and the fuzzing test.

Moves towards an approach where the service layer has multiple "build" methods which can build vega protos for submitting transactions.

In a future PR this can be applied to existing commands and tidied up into it's own module which the service layer can access through compartmentalisation. Basic service layer methods for doing 'frequent' commands like submit order or submit liquidity will still exist. More complicated orders like a pegged iceberg order will not be supported on the service layer.

Approach prevents a cluttered service layer with methods with 30+ fields.

### Testing
Currently failing fuzz test - blocked by https://github.com/vegaprotocol/vega/issues/9811

### Breaking Changes
- `create` methods renamed to `build`
- moved `build_pegged_order` to service layer in preparation to be moved to new module


vega_sim/api/data.py
```diff
@@ -397,6 +397,144 @@ class TeamRefereeHistory:
...
+        if stop_order.updated_at != 0
+        else None,
+        order_id=stop_order.order_id,
+        party_id=stop_order.party_id,
+        market_id=stop_order.market_id,
+        rejection_reason=stop_order.rejection_reason,
+        price=num_from_padded_int(stop_order.price, decimal_spec.price_decimals)
+        if stop_order.price is not None
+        else None,
+        trailing_percent_offset=stop_order.trailing_percent_offset
```
TomMcL: Does this come in as a float or does it need converting from a string?

TomMcL (commented):
Looks good, one small query

vega_sim/api/data.py
cdummett: If the docs are correct, it should be decimal value:

> This should be expressed as a decimal value between 0 and 1, e.g. 0.01 for 1%

Which I think is the right format for us? Unless you think it is misleading having the word percent but using a float?

TomMcL (approved):
lgtm

PR #523, opened 2023-10-10
feat: new spam test 

Sohill-Patel:
### Description

Spam test for AC

### Closes

close https://github.com/vegaprotocol/OctoberACs/issues/390
close https://github.com/vegaprotocol/OctoberACs/issues/392
close https://github.com/vegaprotocol/OctoberACs/issues/394


davidsiska-vega (changes requested):
Even if we accept that null chain is enough to test spam:
The actual system test, testing the AC should live in the system test repo. So in the system test repo you can import the market sim package, run the tests you want on top of market sim / null chain. 

This ensures that 
a) the test is run as mandatory on PRs / included in overnight runs (as you set it)
b) market sim codebase is kept for it's purpose - which is market simulations.

Sohill-Patel:
```{“vega":"9755-spam-stats-extension"}```

**Comment by Sohill-Patel** [comment]
```{"vega":"develop"}```

File: .env
Comment (cdummett) [inline .env:1]: This needs to be changed to a specific commit (whatever the recent develop commit is good - as long as it passing).

When you do make sure the protos are up to date with `make proto` 😉

File: Jenkinsfile [hunk 1]
```diff
@@ -14,7 +14,7 @@ pipeline {
         disableConcurrentBuilds(abortPrevious: true)
     }
     parameters {
-        string( name: 'VEGA_VERSION', defaultValue: 'ae2715f513f75351f7f78546ea58ae034667564c',
+        string( name: 'VEGA_VERSION', defaultValue: 'cb2153cc1d6bba5d8ac348ed4b2bf8720c428630',
```
cdummett: Make sure this matches what is in `.env`

cdummett (changes requested):
Two small comments - the tests and mods to the service classes look good though.

Just need to address Davids comments as the tests themselves might need to be moved to system-tests.

Sohill-Patel:
moved to https://github.com/vegaprotocol/system-tests-nbc/pull/4

PR #528, opened 2023-10-14
feat: CFM bots

### LINKED ISSUE #522: Create simple "CFM" bot
The aim is to create a simple CFM bot that will place volume according to the $xy=k$ formula. 

The logic should be the following: 
- work out the volume at each tick between one tick from last traded price to min / max implied by the `lpPriceRange` parameter. 
- at each time step it should cancel all it's old volume and dump new volume on the book. 
- Use $xy = k_b$ for incoming buy orders and $xy = k_s$ for incoming sell orders with $k_b$ and $k_s$ adjusted based on position size, so that "if I am already short buy orders get a bad price from me but sell orders get a good price from me". 

Context: we want to get a feeling for how this: https://github.com/vegaprotocol/specs/issues/1990 would work. 

Follow up: once this is done we'll do a ticket for a Uniswap v3 type curve. One step at a time.

davidsiska-vega:
### Description

A LP bot that places volume on the book following $xy = k$ and sets $k$ to control position. 

Closes #522 . 

Authored by @TomMcL - I think in the link you gave me you've not pressed "create".


vega_sim/scenario/common/agents.py
```diff
@@ -261,13 +279,29 @@ def step(self, vega_state: VegaState):
         buy_vol = self.random_state.poisson(self.buy_intensity) * self.base_order_size
         sell_vol = self.random_state.poisson(self.sell_intensity) * self.base_order_size
 
-        best_bid, best_ask = self.vega.best_prices(self.market_id)
+        # best_bid, best_ask = self.vega.best_prices(self.market_id)
+        buy_vwap = self._execution_price(buy_vol, vega_protos.SIDE_BUY)
+        sell_vwap = self._execution_price(buy_vol, vega_protos.SIDE_SELL)
```
cdummett, line 295:
```suggestion
        sell_vwap = self._execution_price(sell_vol, vega_protos.SIDE_SELL)
```

vega_sim/scenario/constant_function_market/scenario.py
cdummett, line 1: Lots of commented out blocks of code in this file?

Is to switch up the scenario quickly or not intentional?

vega_sim/vegahome/genesis.json
```diff
@@ -179,7 +179,7 @@
       "spam.pow.numberOfPastBlocks": "100",
       "spam.pow.numberOfTxPerBlock": "1000",
       "spam.protection.delegation.min.tokens": "100000000000000000",
-      "spam.protection.max.batchSize": "30",
+      "spam.protection.max.batchSize": "200",
```
cdummett, line 192: Not sure about how I feel moving away from `mainnet` values here.

Maybe a separate genesis for sim works now devops are using market-sim bots?

vega_sim/scenario/constant_function_market/scenario.py
```diff
@@ -0,0 +1,459 @@
...
+        scenario.run_iteration(
+            vega=vega,
+            pause_at_completion=False,
+            log_every_n_steps=10,
+            output_data=True,
+        )
+
+    figs = price_comp_plots()
+    for i, fig in enumerate(figs.values()):
+        fig.savefig(f"./cfm_plots/trading-{i}.jpg")
```
cdummett, line 477: I get an error here, manually adding a `cfm_plots` dir fixed it but may be worth adding a block which creates if it doesn't exist or add the dir it to git.

PR #530, opened 2023-10-18
feat: Adding functionality to update market states

TomMcL:
### Description

Adding calls to service for the functionality to change market state through governance

### Testing

Manually tested, state updates. Added integration test

### Breaking Changes

None


vega_sim/service.py
```diff
@@ -1012,6 +1012,44 @@ def update_network_parameter(
...
 
+    def update_market_state(
+        self,
+        market_id: str,
+        proposal_key: str,
+        market_state: vega_protos.governance.MarketStateUpdateType,
+        price: Optional[str] = None,
+        wallet_name: Optional[str] = None,
+        approve_proposal: bool = True,
+    ):
```
cdummett: I'm not too fussed but I thought we were having optional datetimes for proposal methods now?

vega_sim/service.py
```diff
@@ -1012,6 +1012,44 @@ def update_network_parameter(
...
+            wallet_name=wallet_name,
+        )
+        if approve_proposal:
+            gov.approve_proposal(
+                proposal_id=proposal_id,
+                wallet=self.wallet,
+                key_name=proposal_key,
+                wallet_name=wallet_name,
+            )
+            self.wait_fn(60)
```
cdummett: With `forward_time_to_enactment` functionality instead of the hardcoded waits (depends on above comment though)?

cdummett (changes requested):
Looks good - two small comments and I think the protos are out of date.

cdummett (approved):

PR #535, opened 2023-10-24
Feat/perps

witgaw:
* Add method for creating perps market.
* Allow injecting perps settlement data.
* Map product data within market data.
* Modify `examples.nullchain` to run a perps market (when called with `--perps` flag).
* Target latest `core` commit on develop.


witgaw (commented):
Ready for review

.env
```diff
@@ -1,4 +1,4 @@
-VEGA_SIM_VEGA_TAG=dd2725c0f575486c9a080a7ac92947a30887cbce
+VEGA_SIM_VEGA_TAG=edce663b68815ccd8a1ceb6fd0dffe9af40f86e2
```
cdummett: Will need to update the tag in the `Jenkins` file as well to match this change.

vega_sim/api/data.py
```diff
@@ -1019,6 +1029,30 @@ def _liquidity_provider_fee_share_from_proto(
...
+            ),
+            funding_rate=float(data.funding_rate),
+            internal_twap=num_from_padded_int(
+                data.internal_twap, decimal_spec.price_decimals
+            ),
+            external_twap=num_from_padded_int(
+                data.external_twap, decimal_spec.price_decimals
+            ),
+        )
+    raise Exception(f"unsupported product data type '{data_field}'")
```
cdummett: Nitpick but I would prefer if `_product_data_from_proto` returned a custom `ProductData` dataclass (with a `perpetual_data` field) so the results are structured in the same way as the APIs. Might make it easier if something gets added to `ProductData` in the future.

vega_sim/service.py
```diff
@@ -1232,25 +1389,29 @@ def update_market(
...
         settlement_key: str,
         settlement_price: float,
         market_id: str,
         wallet_name: Optional[str] = None,
     ):
-        future_inst = data_raw.market_info(
-            market_id, data_client=self.trading_data_client_v2
-        ).tradable_instrument.instrument.future
+        product = helpers.get_product(
+            data_raw.market_info(market_id, data_client=self.trading_data_client_v2)
```
cdummett: We should be able to use `self._market_from_feed(market_id)` here to save a request.

vega_sim/service.py
```diff
@@ -1260,6 +1421,41 @@ def settle_market(
...
+    def submit_settlement_data(
+        self,
+        settlement_key: str,
+        settlement_price: float,
+        market_id: str,
+        wallet_name: Optional[str] = None,
+        additional_payload: Optional[dict[str, str]] = None,
+    ):
+        product = helpers.get_product(
+            data_raw.market_info(market_id, data_client=self.trading_data_client_v2)
```
cdummett: Ditto here, we should be able to use `self._market_from_feed(market_id)` here to save a request.

vega_sim/service.py
```diff
@@ -1445,6 +1641,17 @@ def get_latest_market_data(
...
+    def get_market_state(
+        self,
+        market_id: str,
+    ) -> str:
+        """
+        Output market state.
+        """
+        return vega_protos.markets.Market.State.Name(
+            self.get_latest_market_data(market_id).market_state
+        )
```
cdummett: Another nitpick, I think we are trying to work towards the point where service methods expect and return vega enums rather than names.

There are a lot of places where this isn't yet true though so maybe leave for now and we can raise an issue to review everything in the future. Thoughts @TomMcL?

vega_sim/service.py
```diff
@@ -1570,9 +1777,9 @@ def submit_simple_liquidity(
             wallet_name:
                 str, The name of the wallet which is placing the order
         """
-        asset_id = data_raw.market_info(
+        asset_id = data_raw.settlement_asset(
             market_id=market_id, data_client=self.trading_data_client_v2
-        ).tradable_instrument.instrument.future.settlement_asset
```
cdummett: Should be able to use the asset market map here to save a request
```
asset_id = self.market_to_asset[market_id]
```

vega_sim/service.py
```diff
@@ -211,6 +228,17 @@ def market_to_asset(self) -> str:
             )
         return self._market_to_asset
 
+    @property
+    def market_to_asset(self) -> str:
+        if self._market_to_asset is None:
+            self._market_to_asset = DecimalsCache(self.__extract_settlement_asset)
+        return self._market_to_asset
```
cdummett: This is overriding the `market_to_assset` property just above, will need to pick one to keep.

cdummett (commented):
Hey looks great so far, a few small comments

.env
witgaw: Thanks, done now. Pushed to see how CI copes.

vega_sim/service.py
witgaw: Thanks, modified now I think, but please verify.

vega_sim/service.py
witgaw: Thanks, removed now.

vega_sim/service.py
witgaw: Fixed now.

examples/nullchain.py
```diff
@@ -1,11 +1,14 @@
...
 
 from vega_sim.null_service import VegaServiceNull
-import vega_sim.proto.vega as vega_protos
-from vega_sim.proto.vega.governance_pb2 import UpdateMarketConfiguration
+from vega_sim.service import MarketStateUpdateType
 
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-p", "--perps", action="store_true", default=False)
```
TomMcL: This part should be inside the `__name__` part (doesn't hugely matter since it's just an example file but still)

vega_sim/api/data.py
witgaw: Fair enough, modified now.

vega_sim/service.py
witgaw: Wouldn't local enums to translate between the two? Unless I'm missing some python/proto plugin that will give me intellisense for those types.

vega_sim/service.py
```diff
@@ -1232,25 +1389,29 @@ def update_market(
         self.wait_fn(60)
         self.wait_for_thread_catchup()
 
-    def settle_market(
+    def submit_termination_and_settlement_data(
```
TomMcL, line 1381: Will need to tell console-test to update their calls to this function

vega_sim/service.py
TomMcL: Yeah we should be using the enum itself wherever possible. I know we pass it in as strings in a few places (which is more just because of the advantage that the proto can handle that) but we should avoid returning them as strings.

vega_sim/service.py
TomMcL: Yeah I was sure we already had this when I saw it :D

examples/nullchain.py
witgaw: Sure, moved now.

vega_sim/service.py
witgaw: Cheers, sorted now.

witgaw (commented):
Addressed review comments

vega_sim/service.py
TomMcL: @witgaw local enums to translate would probably be the ideal state indeed given that the proto definition isn't handled well by autocomplete, however that would be a pain to manage :D

TomMcL (approved):

cdummett (approved):

PR #550, opened 2023-11-13
feat: faucet startup wait

TomMcL:
### Description

Adding wait between starting node and faucet to avoid case where faucet fails before node is fully started up

### Testing

Integration tests


vega_sim/null_service.py
```diff
@@ -487,6 +487,14 @@ def manage_vega_processes(
         log_name="node",
     )
 
+    for _ in range(500):
+        try:
+            requests.get(
+                f"http://localhost:{port_config[Ports.CORE_REST]}/blockchain/height"
+            ).raise_for_status()
```
daniel1302, line 494: Should We add break if request is successful to avoid further requests if core API is already up and running?

TomMcL: We should yes :D

cdummett (approved):

PR #551, opened 2023-11-14
fIx: fuzzing agents except custom VegaProtoValueError

cdummett:
### Description
All builder methods now catch `ValueError` exceptions and instead raise a custom `VegaProtoValueError`.

This allows fuzzing agents to except this custom exception when fuzzed values cause a value error.

### DIFF
--- a/vega_sim/builders/exceptions.py
+++ b/vega_sim/builders/exceptions.py
@@ -1,16 +1,18 @@
 import logging
 
 from typing import Callable
+from functools import wraps
 
 
 def raise_custom_build_errors(func: Callable):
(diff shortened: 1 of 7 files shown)


vega_sim/builders/exceptions.py
```diff
@@ -0,0 +1,18 @@
+import logging
+
+from typing import Callable
+
+
+def raise_custom_build_errors(func: Callable):
```
TomMcL, line 7: I expect this could do with a functools wrap

TomMcL (approved):

PR #552, opened 2023-11-14
feat: support perps in fuzzing scenario

witgaw:
TODO (to be added in next PR):
* Fuzz funding payments (non-homogenous)
* Fuzz external data sampling (non-homogenous)
* Perp-specific figures


vega_sim/scenario/common/agents.py
```diff
@@ -3528,3 +3528,54 @@ def finalise(self):
...
+        self.current_step += 1
+        self.prev_data = self.curr_data
+        self.curr_data = next(self.oracle_data_generator)
+
+        # depending on order in which agents are traversed the market may not be initialised yet
+        if self.market_id is None:
+            id = self.vega.find_market_id(name=self.market_name)
+            if id is None:
+                return
+            self.market_id = id
```
witgaw: I'm not loving this, but alternative was to just put all of that functionality into the `FuzzySuccessorConfigurableMarketManager` which didn't feel right, but perhaps that's the way to go as it would avoid the problems of:
* the specified  `perp_settlement_key_name` needing to exist before it's specifed in the `FuzzySuccessorConfigurableMarketManager`,
* market not existing when the `SettlementDataSubmitter` gets instantiated.

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -831,9 +848,18 @@ def __init__(
...
+                "'perp_settlement_wallet_name' must be supplied when `market_config` indicated as perp market"
+            )
+
         self.settlement_price = settlement_price
         self.needs_to_update_markets = False
         self.stake_key = stake_key
+        self.perp_options = perp_options
+        self.perp_assure_enabled = perp_assure_enabled
+        self.perp_close_at_settlement_price = perp_close_on_finalise
+        self.perp_settlement_key_name = perp_settlement_key_name
```
witgaw: Should these go into some DTO or is it fine like that?

witgaw (commented):
Ready for initial review

examples/nullchain.py
```diff
@@ -80,24 +80,9 @@
...
-                        "perps market proposals not allowed by default, allowing via network parameter change failed"
-                    )
-                else:
-                    print(
-                        f"successfully updated network parameter '{perps_netparam}' to '{new_val}'"
-                    )
-
+            vega.try_enable_perp_markets(
+                proposal_key=MM_WALLET.name, raise_on_failure=True
+            )
```
TomMcL: Probably we should just enable this in the genesis file and not have to worry about it

vega_sim/api/market.py
```diff
@@ -169,6 +192,12 @@ def build(self):
     def set(self, parameter, value):
         rsetattr(self, attr=parameter, val=value)
 
+    def is_future(self) -> bool:
+        return self.instrument.future != None
```
TomMcL:
```suggestion
        return self.instrument.future is not None
```

vega_sim/api/market.py
```diff
@@ -169,6 +192,12 @@ def build(self):
...
+
+    def is_perp(self) -> bool:
+        return self.instrument.perp != None
```
TomMcL:
```suggestion
        return self.instrument.perp is not None
```

http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

vega_sim/api/market.py
```diff
@@ -361,20 +390,40 @@ class InstrumentConfiguration(Config):
...
         self.name = config["name"]
         self.code = config["code"]
-        self.future = FutureProduct(opt=config["future"])
+        self.future = None
+        future_config = config["future"]
+        if future_config != None:
+            self.future = FutureProduct(opt=future_config)
+        self.perp = None
+        perp_config = config["perp"]
+        if perp_config != None:
```
TomMcL: Could compress these into `if config["perp"] is not None` (or even a walrus operator if you feel like it)

vega_sim/scenario/common/agents.py
TomMcL: Wouldn't be averse to just merging it all into a MarketManager no. (also, best not to call things in python `id` as it's a built-in function which you'll overwrite)

vega_sim/scenario/common/agents.py
```diff
@@ -3528,3 +3528,54 @@ def finalise(self):
...
+
+        self.vega.submit_settlement_data(
+            settlement_key=self.key_name,
+            settlement_price=self.curr_data,
+            market_id=self.market_id,
+        )
```
TomMcL: Maybe add an argument for how frequently this actually submits so you can play around with that?

vega_sim/scenario/fuzzed_markets/agents.py
TomMcL: Nah this is fine IMO, we're not going to have some explosion in types of markets

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -944,6 +1007,14 @@ def _create_latest_market(self, parent_market_id: Optional[str] = None):
             )
 
         self.vega.wait_for_total_catchup()
+
+        if self.perp_assure_enabled and mkt_config.is_perp():
```
TomMcL: Can drop this if we just enable in genesis

TomMcL (changes requested):

vega_sim/api/market.py
witgaw: Amended now, thanks a lot.

examples/nullchain.py
witgaw: Fair enough, added now.

vega_sim/scenario/fuzzed_markets/agents.py
witgaw: Done now

vega_sim/api/market.py
witgaw: Thanks, done now (I think)

vega_sim/scenario/common/agents.py
witgaw: OK, moved it there

vega_sim/scenario/common/agents.py
witgaw: Good point, added now

TomMcL (approved):

PR #571, opened 2023-11-30
feat: check balances equal deposits

### LINKED ISSUE #570: Add balances invariance checks to market - ideally all runs
Add balance invariance checks. So that market sim keeps track of:
1. all monies fauceted across time
2. at the end of run sum up all the balances in all vega accounts 

Verify we have those two amounts equal.

cdummett:
PR adds a method for checking balances across all accounts match the total amount minted for each asset.

PR also adds a `--lite` flag to the fuzzing script so new features can be tested against the fuzzing scenario faster. 

closes #570


davidsiska-vega (dismissed):
Looking good!

davidsiska-vega (dismissed):

TomMcL (approved):

PR #572, opened 2023-11-30
feat: auction exit visualisation

cdummett:
### Description
PR adds an `AuctionExitVisualisation` visualisation which shows a market entering and exiting a price monitoring auction.

Example command for running visualisation with console and break points:
```
python -m examples.visualisations.auction_exit --console --pause
```

In CI parties trading in auctions have their trades checked to ensure they incur no fees.

### DIFF
--- a/examples/visualisations/utils.py
+++ b/examples/visualisations/utils.py
@@ -88,7 +88,9 @@ def continuous_market(
     asset_id = propose_asset(vega=vega)
     mint_settlement_asset(vega=vega, asset_id=asset_id)
 
-    market_id = propose_market(vega=vega, asset_id=asset_id, price_monitoring=price_monitoring)
+    market_id = propose_market(
+        vega=vega, asset_id=asset_id, price_monitoring=price_monitoring
+    )
     provide_liquidity(vega=vega, market_id=market_id, fee=liquidity_fee)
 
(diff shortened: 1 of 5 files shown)


examples/visualisations/utils.py
```diff
@@ -194,28 +217,26 @@ def propose_market(
...
-    vega.create_simple_market(
-        market_name="XYZ:DAI Visualisation Example",
-        wallet_name=AUX_PARTY_A.wallet_name,
-        proposal_key=AUX_PARTY_A.key_name,
-        termination_wallet_name=AUX_PARTY_A.wallet_name,
-        termination_key=AUX_PARTY_A.key_name,
-        settlement_asset_id=asset_id,
-        market_decimals=3,
+    if not price_monitoring:
+        market_config.set("price_monitoring_parameters.triggers", [])
```
TomMcL: Does this work ok? I remember in the past a market with no price monitoring parameters at all did not really behave well.

cdummett: Seems to work okay but have realised my reasoning for doing it is flawed anyway...

Thought `create_simple_market` which this was replacing didn't use price monitoring.. will update it to set the current defaults.

TomMcL (approved):

PR #573, opened 2023-12-05
feat: minting imrovements

### LINKED ISSUE #564: Faucet use improvements
Instead of calling the faucet many times for different agents for various small(ish) amounts and then waiting let's try:
1) set up a "treasurer" or "marketSimFaucet" party (not sure what's the best name). 
2) this party will call the faucet once (on each asset) for the maximum possible amount (maxInt?). 
3) we will use internal vega transfers instead to fund individual agents as and when needed (this will need us to max out the setting on spam on maximum transfers per block).

cdummett:
PR adds a "treasury" key/wallet to the base `VegaService`.

- On startup the staking asset (symbol `VOTE`) is minted from the faucet to the key/wallet.
- On asset creation the asset is minted from the faucet to the key/wallet.

The mint method is then updated with a `from_faucet` arg (defaults to `False`). If not minting from the faucet, the key calling the method will be topped up with a transfer from the "treasury".

closes https://github.com/vegaprotocol/vega-market-sim/issues/564


vega_sim/service.py
```diff
@@ -512,6 +531,18 @@ def create_asset(
         self.wait_fn(60)
         self.wait_for_thread_catchup()
 
+        asset_id = self.find_asset_id(
+            symbol=symbol, enabled=True, raise_on_missing=True
+        )
+        for _ in range(20):
```
TomMcL: Should we not just raise the max faucet amount rather than doing it 20 times, or is it so fast as not to matter? :)

cdummett: Yeah fair point this was a quick fix that avoiding updating every scenario but agree it would be cleaner, will update it now 👍

TomMcL (approved):

PR #580, opened 2023-12-08
fix: retry faucet calls

cdummett:
### Description
CI crashing after the faucets attempted connection to the core API is rejected.

```
[2023-12-08T09:14:15.719Z] ERROR    vega_sim.api.faucet:faucet.py:20 Exception in minting: {'error': 'rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:38665: connect: connection refused"'}
```

PR adds an exponential wait between faucet calls to investigate if core will eventually accept the faucet calls. The total wait time before raising an exception is `~=18.7s`

### DIFF
--- a/vega_sim/api/faucet.py
+++ b/vega_sim/api/faucet.py
@@ -1,3 +1,4 @@
+import time
 import requests
 from logging import getLogger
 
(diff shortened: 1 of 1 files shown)


davidsiska-vega (approved):
Let's try.

PR #590, opened 2024-01-10
feat: Adding default mark price config

TomMcL:
### Description

Changes to mark price specification for core. Default should act as before

### Testing

Integration tests


davidsiska-vega (approved):
Purely based on running integration tests locally and some of the fuzzing (300 steps). 

If you want an actual code review... @cdummett or someone else will have to step in.

PR #594, opened 2024-01-17
feat: add network pnl plots

cdummett:


cdummett:
Closing as datanode pnl fixes no longer planned.

PR #601, opened 2024-01-29
feat: improved fuzzed traders

### LINKED ISSUE #586: Fuzzing: Widen range of agents fuzzed values
Currently most agent values are sampled from a Poisson distribution when fuzzing. Enhance the market sim fuzzing run such that:

 - Agents sample across the range whilst also mostly trading in sensible ranges so some trading happens
 - Agents deliberately submit near upper/lower bounds of values

cdummett:
Improvements:
- random probability trader places order at bounds using beta distribution
- proportion of submissions, amendments, cancellations and stop orders moved outside batch instruction
- fuzzing function with validity bias added so more "transactions are accepted by core

closes https://github.com/vegaprotocol/vega-market-sim/issues/586

### DIFF
--- a/vega_sim/scenario/fuzzed_markets/agents.py
+++ b/vega_sim/scenario/fuzzed_markets/agents.py
@@ -32,23 +32,6 @@ import vega_sim.scenario.fuzzed_markets.fuzzers as fuzzers
 class FuzzingAgent(StateAgentWithWallet):
     NAME_BASE = "fuzzing_agent"
 
-    # Set the memory which all instances can modify
-    MEMORY = {
-        key: []
-        for key in (
-            "TRADING_MODE",
-            "COMMAND",
-            "TYPE",
-            "SIDE",
-            "TIME_IN_FORCE",
-            "PEGGED_REFERENCE",
-            "POST_ONLY",
-            "REDUCE_ONLY",
-        )
-    }
-    # Set the output flag which all instances can modify
-    OUTPUTTED = False
-
     def __init__(
         self,
         key_name: str,
@@ -59,7 +42,6 @@ class FuzzingAgent(StateAgentWithWallet):
         state_update_freq: Optional[int] = None,
         initial_asset_mint: float = 1e9,
         random_state: Optional[RandomState] = None,
-        output_plot_on_finalise: bool = False,
     ):
         super().__init__(
             key_name=key_name,
@@ -72,7 +54,6 @@ class FuzzingAgent(StateAgentWithWallet):
         self.asset_name = asset_name
         self.initial_asset_mint = initial_asset_mint
         self.random_state = random_state if random_state is not None else RandomState()
-        self.output_plot_on_finalise = output_plot_on_finalise
 
     def initialise(
         self, vega: VegaServiceNull, create_key: bool = True, mint_key: bool = True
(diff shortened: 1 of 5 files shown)


vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -105,22 +86,93 @@ def step(self, vega_state):
...
+                    self.vega.submit_transaction(
+                        key_name=self.key_name,
+                        wallet_name=self.wallet_name,
+                        transaction=cancellation,
+                        transaction_type="order_cancellation",
+                    )
+                except HTTPError:
+                    continue
+        stop_orders_submissions = []
+        for _ in range(0):
```
TomMcL: You know this just means it won't trigger right?

vega_sim/scenario/fuzzed_markets/agents.py
```diff
@@ -142,150 +194,25 @@ def create_fuzzed_cancellation(self, vega_state):
 
         return self.vega.build_order_cancellation(
             order_id=order_id,
-            market_id=self.market_id,
+            # market_id=self.market_id,
```
TomMcL, line 197: No more market id?

vega_sim/scenario/fuzzed_markets/fuzzers.py
```diff
@@ -527,3 +527,219 @@ def _pick_recurring():
...
+                    peak_size=rs.uniform(0, 1.1) * size,
+                    minimum_visible_size=rs.uniform(0, 0.6) * size,
+                ),
+                None,
+            ],
+            p=[0.1, 0.9],
+        )
+
+    # Get network information
+    market_ids = [key for key, _ in vega.market_to_asset.items()]
```
TomMcL:
```suggestion
    market_ids = list(vega.market_to_asset.keys())
```

vega_sim/scenario/fuzzed_markets/fuzzers.py
```diff
@@ -527,3 +527,219 @@ def _pick_recurring():
...
+
+def fuzz_order_submission(
+    vega: VegaService, rs: RandomState, bias: float
+) -> vega_protos.commands.v1.commands.OrderSubmission:
+    def _pick_market_id():
+        if len(market_ids) == 0:
+            return None
+        return rs.choice(market_ids)
+
+    def _pick_type():
```
TomMcL, line 540: Don't call variables `type` :). `typ` / `typez` / `type_` could be options

vega_sim/scenario/fuzzed_markets/fuzzers.py
```diff
@@ -527,3 +527,219 @@ def _pick_recurring():
...
+    market_ids = [key for key, _ in vega.market_to_asset.items()]
+    market_id = _pick_market_id()
+    market_data = vega.market_data_from_feed(market_id)
+
+    # Pick driver fields
+    type = _pick_type()
+    time_in_force = _pick_time_in_force()
+    pegged_order = _pick_pegged_order()
+
+    # Pick driven fields
```
TomMcL, line 655: In future I'd probably have leant towards these all being top level functions and actually taking in the fields they wanted to make it a bit clearer what's going on here! But I'd leave as it is for now

vega_sim/scenario/fuzzed_markets/agents.py
cdummett: Was cancelling all orders on the market too often when an order id was not specified (sometimes the feed lags core).

Could set it to sometimes specify the market id and sometimes not so we still test this feature? For the most part I just want to ensure we have lots of persistent orders on the book for each agent to amend since that seems to be the area we were lacking on before.

vega_sim/scenario/fuzzed_markets/agents.py
cdummett: Thanks testing typo. Will update.

vega_sim/scenario/fuzzed_markets/fuzzers.py
cdummett: Sounds good, happy to rework them to do that next time working on this area 👍

TomMcL (approved):
lgtm, only test breaking on known core bug

Issue #587, opened 2024-01-04
Fuzzing: don't leave opening auction immediately...


davidsiska-vega:
... instead spend some time doing all the fuzzing stuff while the market is still in opening auction.

PR #604, opened 2024-01-31
feat: Don't immediately leave opening auction in fuzz tests

TomMcL:
### Description

Adding flag to opening auction exit agents to not always immediately exit

### Testing

Ran locally, observed that auction stuck around for a few steps

### Breaking Changes

None, default behaviour as before

### Closes

#587


davidsiska-vega (approved):

cdummett:
@TomMcL There are some `LimitOrderTrader` agents in the fuzzing scenario who place limit orders distributed pretty close to the reference price as well. Might need to add some logic to those as well so they don't place orders during opening auctions (or do the same probability thing)?

PR #605, opened 2024-01-31
feat: default to last success

cdummett:
If the fuzzed market proposer is unable to propose a market successfully, currently it proposes the market-sim default market. Since there is already a separate chain of default sensible markets this adds no value.

Instead, it should propose the last successful fuzzed config.


TomMcL (approved):
LGTM

PR #617, opened 2024-02-23
feat: Add batch proposals

TomMcL:
### Description

Adding batch governance proposals

### Testing

Swapped one fuzz tester to use

### DIFF
--- a/vega_sim/builders/governance.py
+++ b/vega_sim/builders/governance.py
@@ -12,7 +12,7 @@ import datetime
 
 import vega_sim.proto.vega as vega_protos
 
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, Union
 from vega_sim.api.helpers import num_to_padded_int
 from vega_sim.builders.exceptions import raise_custom_build_errors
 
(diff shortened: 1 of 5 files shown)


davidsiska-vega (dismissed):

vega_sim/builders/governance.py
```diff
@@ -44,14 +44,27 @@ def proposal_terms(
...
+    for_batch_proposal: bool = False,
+) -> Union[
+    vega_protos.governance.ProposalTerms,
+    vega_protos.governance.BatchProposalTermsChange,
+]:
+    if not for_batch_proposal:
+        if closing_timestamp is None:
+            raise Exception(
+                "For non batch proposals changes closing timestamp must be set"
+            )
```
cdummett: Could we not have this check here?

I was thinking these build methods would be "dumb" so the fuzzers can try wrong things without errors being raised and us having to catch them else where (the exception being the `VegaProtoValueError`s raised by the wrapper).

Could have this sort of validation in any service methods that use the builder instead?

TomMcL: Sure, makes sense, let's kill it

cdummett (approved):

PR #618, opened 2024-02-25
Add handling for funding period events and oracle data

witgaw:
+ notebook for data download


vega_sim/api/data.py
```diff
@@ -304,12 +305,26 @@ class LiquidityProviderFeeShare:
     average_score: float
 
 
+@dataclass(frozen=True)
+class LiquiditySLA:
```
cdummett: Small nitpick but could we have this dataclass match the corresponding proto `LiquidityProviderSLA`?

vega_sim/api/data.py
```diff
@@ -282,8 +282,9 @@ class MarketData:
     market_value_proxy: float
     price_monitoring_bounds: list
     liquidity_provider_fee_share: list
+    liquidity_sla: list
```
cdummett: Would give some useful type hinting when working with the `MarketData` dataclass.
```suggestion
    liquidity_sla: List[LiquidityProviderSLA]
```

vega_sim/api/data.py
```diff
@@ -1053,7 +1123,7 @@ def positions_by_market(
     asset_decimals_map: Optional[Dict[str, int]] = None,
 ) -> Union[Dict[str, Position], Position]:
     """Output positions of a party."""
-
+    data_raw
```
cdummett:
```suggestion
```

vega_sim/api/data.py
```diff
@@ -1245,6 +1334,73 @@ def _liquidity_provider_fee_share_from_proto(
...
+                for x in individual_liquidity_provider_sla.hysteresis_period_fee_penalties
+            ],
+            required_liquidity=(
+                0
+                if individual_liquidity_provider_sla.required_liquidity == ""
+                else num_from_padded_int(
+                    float(individual_liquidity_provider_sla.required_liquidity),
+                    decimals,
+                )
+            ),
```
cdummett, line 1381: I'm not 100% but am trying to check now. Are the right decimals being used here?

I would have assumed it was just the asset decimals.

vega_sim/api/data.py
witgaw: Renamed now

vega_sim/api/data.py
witgaw: Yeah, after a bit of digging it does seem to be asset decimals, changed now.

cdummett (approved):

PR #627, opened 2024-03-11
feat: allow tick sizes greater than 1

cdummett:
Updates all service methods which submit/amend orders at prices to have an optional round to tick field. When true..
- prices of buy orders will be rounded down to nearest tick
- prices of sell orders will be rounded up to nearest tick

The tick size field of all existing configs is also updated so the feature is being covered.

### DIFF
--- a/vega_sim/service.py
+++ b/vega_sim/service.py
@@ -985,6 +985,7 @@ class VegaService(ABC):
         post_only: bool = False,
         peak_size: Optional[float] = None,
         minimum_visible_size: Optional[float] = None,
+        round_to_tick: bool = True,
     ) -> Optional[str]:
         """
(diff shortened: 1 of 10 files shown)


vega_sim/service.py
```diff
@@ -1057,6 +1058,14 @@ def submit_order(
             if price is not None
             else None
         )
+        if round_to_tick and price is not None:
+            # TODO: Implement tick size into the local data cache to avoid API calls
```
TomMcL: Could we at least have a dict here to cache it before it's fully integrated into the data cache? I think calling this on every order might slow things down a fair bit. I guess just blindly clear the cache if update market is called in case it's being changed.

TomMcL (approved):

PR #636, opened 2024-03-15
feat: add retry check

cdummett:
To ensure failed inconsistent market data checks are not a result of out of sync nodes still processing blocks, if a check fails, we wait a significant period and then retry the check.

### DIFF
--- a/vega_sim/service.py
+++ b/vega_sim/service.py
@@ -3645,18 +3645,30 @@ class VegaService(ABC):
             margin_factor=str(margin_factor),
         )
 
-    def check_market_states_consistent(self, raise_exceptions: bool = True):
+    def check_market_states_consistent(
+        self, raise_exceptions: bool = True, max_attempts: int = 5
+    ):
+        exception = None
         self.wait_for_total_catchup()
-        markets = self.all_markets()
-        for market in markets:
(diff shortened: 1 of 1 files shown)


vega_sim/service.py
```diff
@@ -3645,18 +3645,30 @@ def update_margin_mode(
...
+                    assert market.trading_mode == market_data.market_trading_mode
+                    return
+                except AssertionError as e:
+                    exception = e
+                    time.sleep(5)
+                    logging.warning(
+                        "Check failed, waiting to ensure sync then retrying check."
+                    )
+                    self.wait_for_total_catchup()
+                    continue
```
TomMcL: I think you want a `break` here?

vega_sim/service.py
```diff
@@ -3645,18 +3645,30 @@ def update_margin_mode(
...
+        if raise_exceptions:
+            raise exception
+        logging.error(exception)
```
TomMcL: Best to have logging done through a `logger` defined with `get_logger` at the top of the module

TomMcL (changes requested):

TomMcL (approved):
lgtm

PR #644, opened 2024-03-27
chore: set blockchains.evmChainConfig so netparam propagation doesn't…

wwestgarth:
So that they don't fall over with the second-bridge changes.


wwestgarth:
```
{
  "vega": "develop",
  "vegacapsule": "develop"
}
```

EVODelavega (approved):

witgaw (approved):

PR #646, opened 2024-03-28
fix: covert padded ints to decimals

cdummett:
Converts padded ints to decimals rather than floats to avoid rounding issue.

Sneaks in a fix for replays where we recreate create the governance asset which was causing an app hash mismatch and also defaults store transactions to false when replaying to avoid overwriting the replay file.


vega_sim/null_service.py
```diff
@@ -1006,9 +1006,6 @@ def start(self, block_on_startup: bool = True) -> None:
             time.sleep(6)
             self.process_pids = parent_conn.recv()
 
-        # Initialise the data-cache
-        self.data_cache
-
```
TomMcL, line 1011: ?

cdummett: Just moved too after the early exit if replaying.

Couldn't think of a reason we would need it if replaying?

TomMcL (approved):

PR #653, opened 2024-04-09
feat: implement multiple configurable markets into benchmark scenarios

cdummett:
### Summary:

To allow multiple markets to be run in the same benchmark scenario, the `MarketConfig` class is improved to allow configurable data sources. When building a config from a JSON pulled from a mainnet proposal, data sources in the instrument and composite mark price configuration will now be respected instead of defaulting.

This also allows a market relying on more than one oracle to be tested. The `ConfigurableMarketManager` has been updated to support submitting oracle data for all the relevant oracles of a market.

**NOTE: Since Ethereum oracles are not possible in market-sim without running an emulator like ganache, Ethereum oracles will be converted into standard oracles with any filters respected.**

e.g. this `ethOracle` is automatically converted into this standard `oracle`
```
"ethOracle": {
    "address": "0x719abd606155442c21b7d561426d42bd0e40a776",
    "abi": '[{"inputs": [{"internalType": "bytes32", "name": "id", "type": "bytes32"}], "name": "getPrice", "outputs": [{"internalType": "int256", "name": "", "type": "int256" }], "stateMutability": "view", "type": "function"}]',
    "method": "getPrice",
    "args": ["5i32yLSoX+GmfbRNwS3l2zMPesZrctxliv7fD0pBW0M="],
    "trigger": {
        "timeTrigger": {"initial": "1709575200", "every": "60"}
    },
    "requiredConfirmations": "3",
    "filters": [
        {
            "key": {
                "name": "btc.price",
                "type": "TYPE_INTEGER",
                "numberDecimalPlaces": "18",
            },
            "conditions": [
                {
                    "operator": "OPERATOR_GREATER_THAN",
                    "value": "0",
                }
            ],
        }
    ],
    "normalisers": [
        {"name": "btc.price", "expression": "$[0]"}
    ],
    "sourceChainId": "100",
    }
```
```
"oracle" {
          "signers": {
            "pubKey": {
              "key": "092c117cbcb1be7c125ce5382400d169cb75aeb546e00e0d4780894e0e52aea1"
            }
          }
          "filters" {
            "key": {
              "name": "btc.price"
              "type": "TYPE_INTEGER"
              "numberDecimalPlaces": "18"
            },
            "conditions": [
              {
                "operator": "OPERATOR_GREATER_THAN",
                 "value": "0",
              }
            ],
          }
        }
      }
```

### Examples:

Run a full simulation of all the markets currently on mainnet. The volatility of the price process and the average notional trade volume of each market is individually configured in the registry. 

```
python -m vega_sim.scenario.benchmark.run -m mainnet -s 600 -o -p
```

If using vegapy to check the composite mark price sources, we can see multiple oracles feeding the mark price.

multioracle


vega_sim/api/market.py
```diff
@@ -662,14 +700,16 @@ def load(self, opt: Optional[str] = None):
         )
         self.spot = None if config["spot"] is None else SpotProduct(opt=config["spot"])
 
-    def build(self):
+    def build(self, pubkey: str):
```
TomMcL: Could we rename `pubkey` in all of these to something like `oracle_pubkey` to make it clearer what it's being used for? (And could it not be set through the same avenues as we currently set values rather than adding an argument?)

vega_sim/scenario/benchmark/run.py
```diff
@@ -57,79 +60,94 @@ def _run(
...
-            )
-        if is_perpetual:
-            funding_periods = service.api.data.list_funding_periods(
-                market_id=market.id,
-                start_timestamp=market.market_timestamps.pending,
-            )
-            funding_period_data_points = (
-                service.api.data.list_funding_period_data_points(
-                    market.id, start_timestamp=market.market_timestamps.pending
+            # Initialise figures
```
TomMcL: It doesn't look like we ever actually use fig1/2/3 other than for saving them separately. Should we just do away with the 1/2/3 stuff and immediately create + save if output is true?

vega_sim/api/market.py
cdummett: `pubkey` -> `oracle_pubkey` sounds good will make that change 👍 

The logic to pass the pubkey in rather than set in the config started was...

- the `ethOracle` message doesn't have a `pubkey` field to set, but will eventually need one when we convert it to a standard `oracle` message. I wanted the market config to accurately represent the actual proposal and only make the necessary mappings when the proto was built hence the pubkey is only passed in when building.
- for some configurations there are a lot of oracles to set, the `ConfigurableMarketManager` would have to iterate over the config and do a lot of conditional checks to set a public key for each oracle. That logic also then has to get copied over to any other agents or users that want to propose a market from a config.

It's not perfect though because the main functionality we lose is what if we want a market's oracles to be submitted by different keys... seem's niche enough to ignore though.

vega_sim/scenario/benchmark/run.py
cdummett: Good shout thanks, that is much cleaner.

TomMcL (approved):

PR #665, opened 2024-04-29
chore: multi trigger auction configs

cdummett:
### Description
Updates the mainnet BTCUSDT and ETHUSDT market configs to use multiple "short" auction triggers with the same look-back horizon.

For now, existing auctions have been split into four smaller auctions.

(Also updates research configs so overnight fuzzing runs will use multi-trigger auctions)

### Examples

To run a sim of historic prices from a specific date, add the `--datetime` flag

```bash
python -m vega_sim.scenario.benchmark.run -m mainnet-BTCUSDT --steps 1440 --interval 15 -o -p --datetime 2024-04-13:18:00:00
```
price_monitoring_analysis


davidsiska-vega (dismissed):
We may wish to only merge if these go through via governance.

PR #674, opened 2024-05-16
fix: improving minting flow

cdummett:
## Summary 

Current minting method is flaky as there is no guarantee the balance after minting will be greater than the balance before minting, even if the mint is successful. This is as minting directly from the faucet or through transfers requires blocks to be forwarded in which funds can flow out for reasons independent of any flows in. 

Instead...
- we can faucet to the treasury when needed as the treasury balance should be fairly static (i.e. no flows out other than top-up transfers).
- and top-up parties through transfers but crucially check the status of the transfer rather than the account balance.

New minting flow is as follows….

1. Check treasury has enough funds to cover the full “top up” transfer, if not iteratively...
    1. Faucet maximum funds directly to the treasury. 
    1. If funds never arrived raise error.
1. Make an internal “top up” transfer from the treasury to the party with a unique reference. Then..
    1. Forward a block
    1. Wait for synchronisation
    1. Query transfers, checking if transfer is successful. Note we do not care if the balance increased, only that the top up was successful.
    1. If the transfer was rejected raise error.
1. If transfer never appeared raise error.

### DIFF
--- a/vega_sim/local_data_cache.py
+++ b/vega_sim/local_data_cache.py
@@ -453,14 +453,12 @@ class LocalDataCache:
     def initialise_transfer_monitoring(
         self,
     ):
-        base_transfers = []
-
-        base_transfers.extend(
-            data.list_transfers(data_client=self._trading_data_client)
-        )
+        base_transfers = data.list_transfers(data_client=self._trading_data_client)
 
         with self.transfers_lock:
             for t in base_transfers:
+                if t.status != events_protos.Transfer.Status.STATUS_REJECTED:
+                    continue
                 self._transfer_state_from_feed.setdefault(t.party_to, {})[t.id] = t
 
     def initialise_network_parameters(self):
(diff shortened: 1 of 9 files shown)


vega_sim/local_data_cache.py
```diff
@@ -453,14 +453,12 @@ def initialise_market_data(
...
-        base_transfers = []
-
-        base_transfers.extend(
-            data.list_transfers(data_client=self._trading_data_client)
-        )
+        base_transfers = data.list_transfers(data_client=self._trading_data_client)
 
         with self.transfers_lock:
             for t in base_transfers:
+                if t.status != events_protos.Transfer.Status.STATUS_REJECTED:
```
TomMcL: Should this be == ?

TomMcL: Or maybe gone entirely if it was for debugging

cdummett: Yes thanks! Should be `==`.

Thinking way back I think it was so we didn't fill the cache with rejected transfers we didn't care about. Part of me thinks we should scrap transfers from the cache entirely though but I want to do some profiling at some point anyway so will look at that then.

TomMcL (dismissed):

TomMcL (approved):

PR #701, opened 2024-08-30
feat: make drift term available

Jiajia-Cui:


vega_sim/scenario/amm/registry.py
```diff
@@ -14,6 +14,7 @@
                 annualised_volatility=0.5,
                 notional_trade_volume=100,
                 process_theta=0.02,
+                drift=0.02,
```
cdummett: Maybe `process_drift` instead so we know it refers to the price process in the benchmark config?

At the moment it could refer to a drift in any of the other terms.

cdummett (changes requested):
Can we pass this field through into the price process generator in this PR as well please.

At the moment this gives the illusion varying `drift` will modify the underlying OU price process but actually it will still be defaulting to zero.

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+    price_decimals: int,
+    size_decimals: int,
+):
+    x = []  # List to store timestamps
+    y = []  # List to store cumulative volume
+    cumulative_volume = 0  # Initialize cumulative volume
+
+    trades = reversed(trades)
+
+    for trade in trades:
```
cdummett:
```suggestion
    for trade in reversed(trades):
```

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+        # Ensure price and size are available
+        if hasattr(trade, 'price') and hasattr(trade, 'size'):
+            price = padded_int_to_float(trade.price, price_decimals)
+            size = padded_int_to_float(trade.size, size_decimals)
+            
+            # Calculate traded volume (price * size)
+            volume = price * size if price != 0 else 0
+            
+            # Debugging: Print current values
+            print(f"Trade timestamp: {trade.timestamp}, Price: {price}, Size: {size}, Volume: {volume}")
```
cdummett: Clearing up left over prints.

If you would like to leave in debug messages we can add a logger with a logging message here instead.

```suggestion
```

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+            
+            # Accumulate volume (ensure volume is positive)
+            cumulative_volume += abs(volume)
+            y.append(cumulative_volume)
+        else:
+            # If price or size is missing, assume no change in cumulative volume
+            y.append(cumulative_volume)
+            print(f"Missing price or size for trade at {trade.timestamp}")
```
cdummett: Clearing up left over prints.
```suggestion
```

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+
+    # Plot the cumulative volume data using step plot
+    ax.step(x, y, label="Cumulative Volume", where="post")
+
+    # Optionally, set labels and title for clarity
+    ax.set_xlabel('Timestamp')
```
cdummett: Lets leave this out just so we can keep the figure slightly clearer. If we are using datetimes rather than timestamps it should be clear enough what the axis is.

```suggestion
```

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+    ax.set_ylabel('Cumulative Volume')
+    ax.set_title('Cumulative Volume Over Time')
+    ax.legend()
```
cdummett: Overlay methods should just "overlay" data rather than manipulate the axes labels and titles. Could we remove this section please?

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
         y.append(price * size if price != 0 else np.nan)
     ax.step(x, y, label="volume", where="post")
 
+def overlay_cumulative_volume(
+    ax: Axes,
+    trades: List[protos.vega.vega.Trade],
+    price_decimals: int,
+    size_decimals: int,
```
cdummett, line 902: Small change that some (but not all) of the overlays do. Lets you pass mpl kwargs to modify the line plot.

```suggestion
    size_decimals: int,
    **kwargs,
```

vega_query/visualisations/overlay.py
```diff
@@ -894,6 +893,50 @@ def overlay_volume(
...
+            # Accumulate volume (ensure volume is positive)
+            cumulative_volume += abs(volume)
+            y.append(cumulative_volume)
+        else:
+            # If price or size is missing, assume no change in cumulative volume
+            y.append(cumulative_volume)
+            print(f"Missing price or size for trade at {trade.timestamp}")
+
+    # Plot the cumulative volume data using step plot
+    ax.step(x, y, label="Cumulative Volume", where="post")
```
cdummett: Related to above, can unpack the kwargs here.
```suggestion
    ax.step(x, y, label="Cumulative Volume", where="post", **kwargs)
```

cdummett (changes requested):

cdummett (approved):

